PR #254, opened 2025-01-11
Add benchmarking examples using system peripheral and gpio pin toggling

jerrysxie:


examples/rt685s-evk/src/bin/benchmarking-system-peripheral.rs
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+            let cycles = cortex_m::peripheral::DWT::cycle_count();
+            cortex.DWT.disable_cycle_counter();
+            cortex.DCB.disable_trace();
+            info!("DWT Cycle Counter That took {} cycles!!!", cycles);
+        }
+
+        // But if DWT cycle counter is not available, we can use the systick timer
+        {
```
madeleyneVaca, line 39: should this be an else? instead of just wrapped in {}?

examples/rt685s-evk/src/bin/benchmarking-gpio.rs
```diff
@@ -0,0 +1,53 @@
...
+    );
+
+    let n = 1000;
+
+    // Steal the GPIO register map
+    let gpio = unsafe { embassy_imxrt::pac::Gpio::steal() };
+
+    loop {
+        // Do direct write to toggle the pin
+        gpio.not(1).write(|w| unsafe { w.notp().bits(1 << 0) });
```
madeleyneVaca, line 34: what does this not(1) do exactly?

examples/rt685s-evk/src/bin/benchmarking-system-peripheral.rs
jerrysxie: I was trying to showcase both. Now that you mentioned it if-else seems like a better idea. Fixed.

examples/rt685s-evk/src/bin/benchmarking-gpio.rs
jerrysxie: Toggle register is named `not` in the register map; in this case, `not` register of port 1.

PR #268, opened 2025-01-22
[RFC] Enable CTimer based PWM

sukomath:
Enable CTimer based PWM

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -6,8 +6,9 @@ use embassy_hal_internal::interrupt::InterruptExt;
 use embassy_sync::waitqueue::AtomicWaker;
 use paste::paste;
 
-use crate::clocks::{enable_and_reset, ConfigurableClock};
+use crate::clocks::{enable_and_reset, ClockConfig, ConfigurableClock};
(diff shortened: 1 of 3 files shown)


src/timer.rs
```diff
@@ -375,6 +400,174 @@ macro_rules! impl_instance {
...
+                            ct.mcr().modify(|_, w| w.mr1r().set_bit());
+                        }
+                        TimerChannelNum::Channel2 => {
+
+                            ct.mcr().modify(|_, w| w.mr2r().set_bit());
+
+                        }
+                        TimerChannelNum::Channel3 => {
+
+                            ct.mcr().modify(|_, w| w.mr2r().set_bit());
```
madeleyneVaca, line 521: typo? should be mr3r here for channel 3. You could probably also generate this code in a macro, especially if you take Jerry's suggestions

src/timer.rs
```diff
@@ -375,6 +400,174 @@ macro_rules! impl_instance {
...
+                        }
+
+
+                    }
+                }
+
+                fn pwm_enable() {
+                    // SAFETY: safe as long as CTimer PWM is not used across multiple executors
+                    let ct = unsafe { &*crate::pac::[<Ctimer $n>]::steal() };
```
madeleyneVaca, line 530: could be macro generated to simplify this function

PR #271, opened 2025-01-28
Add minimal support for MIMXRT633S

felipebalbi:
With these changes, we have a chance of running embassy-imxrt on MIMXRT633S.


.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[net]
+git-fetch-with-cli = true
```
madeleyneVaca, line 2: is this for the credentials stuff you added in the yamls?

Cargo.toml
```diff
@@ -34,6 +41,20 @@ time-driver = [
     "dep:embassy-time-queue-utils",
 ]
 
+#! ### Chip selection features
+## MIMXRT685S
+mimxrt685s = ["mimxrt685s-pac", "_mimxrt685s"]
+mimxrt633s = ["mimxrt633s-pac", "_mimxrt633s"]
+
+## Reexport the PAC for the currently enabled chip at `embassy_nrf::pac` (unstable)
```
madeleyneVaca, line 49: remove nrf reference?

.cargo/config.toml
felipebalbi: I need to test without it, now that the repositories are all public.

.github/workflows/check.yml
```diff
@@ -133,19 +146,23 @@ jobs:
...
         with:
           submodules: true
           ref: ${{ matrix.commit }}
+
       - name: Install nightly
         uses: dtolnay/rust-toolchain@nightly
+
       - name: cargo doc
-        run: cargo doc --no-deps --all-features
+        run: cargo doc --no-deps -F mimxrt685s,rt,defmt,time,time-driver,unstable-pac
```
tullom, line 165: should we run a cargo doc check for `mimxrt633s` as well?

src/lib.rs
```diff
@@ -32,79 +40,23 @@ pub mod timer;
 pub mod uart;
 pub mod wwdt;
 
+// This mod MUST go last, so that it sees all the `impl_foo!' macros
+#[cfg_attr(feature = "mimxrt633s", path = "chips/mimxrt633s.rs")]
+#[cfg_attr(feature = "mimxrt685s", path = "chips/mimxrt685s.rs")]
+mod chip;
+
 // Reexports
-pub use adc::AdcChannel;
```
tullom, line 36: This was removed, do we need to add it back?

tullom (changes requested):

src/lib.rs
felipebalbi: done

.github/workflows/check.yml
felipebalbi: done

tullom (approved):

jerrysxie (changes requested):

.github/workflows/check.yml
```diff
@@ -155,44 +182,103 @@ jobs:
     runs-on: ubuntu-latest
     name: ubuntu / stable / features
     needs: commit_list
+
     strategy:
       fail-fast: false
       matrix:
         commit: ${{ fromJSON(needs.commit_list.outputs.commits) }}
+        feature: [ "mimxrt633s", "mimxrt633s,rt", "mimxrt633s,defmt",
```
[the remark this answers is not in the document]
felipebalbi: like it was mentioned in this morning's call, yes. We have to do it manually

.github/workflows/nostd.yml
```diff
@@ -3,28 +3,45 @@
...
       - name: Install stable
         uses: dtolnay/rust-toolchain@stable
+
       - name: rustup target add ${{ matrix.target }}
         run: rustup target add ${{ matrix.target }}
+
+      - name: Show variable
+        run: echo ${{ env.TOKEN }}
+
+      - name: Add Credentials
```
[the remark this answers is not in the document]
felipebalbi: done

.github/workflows/check.yml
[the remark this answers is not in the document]
jerrysxie: Gotcha, I was not present in the morning meeting, but if we have to do it manually, then it has to be then.

jerrysxie (approved):

.github/workflows/check.yml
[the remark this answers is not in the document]
felipebalbi: The reason is that once we support more than one chip, we need to have one and only one of them enabled. `cargo hack` will always generate the powerset of features, meaning it will include options where both are enabled and options where none are enabled.

A HAL with more than PAC or no PAC at all is not going to work :-(

Sadly, there's no other way as of now.

jerrysxie:
@madeleyneVaca @JamesHuard Can you guys take a look at this PR when you get a chance?

JamesHuard (approved):

PR #272, opened 2025-01-28
ADC: Fix bit shift in ADC conversion

jerrysxie:
image


src/adc.rs
```diff
@@ -293,7 +295,7 @@ impl<'p, const N: usize> Adc<'p, N> {
         .await;
 
         for e in buf {
-            *e = self.info.regs.resfifo().read().d().bits() as i16;
+            *e = (self.info.regs.resfifo().read().d().bits() >> ADC_BITSHIFT) as i16;
```
felipebalbi, line 298: shouldn't we patch the PAC, instead? Actually, both of them, 685 and 633.

jerrysxie: That would be another way to do it by shortening the bitfield to not include the lower 3 bits. I don't have a strong opinion on this either way.

felipebalbi (approved):

madeleyneVaca (approved):

PR #273, opened 2025-01-29
RFC: Uart Drop() implementation

akshay-rust:
Some of the relevant comments from  https://github.com/OpenDevicePartnership/embassy-imxrt/pull/264  are handled/ resolved here

### DIFF
--- a/src/uart.rs
+++ b/src/uart.rs
@@ -135,11 +135,75 @@ pub type Result<T> = core::result::Result<T, Error>;
 
 impl<'a, M: Mode> UartTx<'a, M> {
     fn new_inner<T: Instance>(_tx_dma: Option<Channel<'a>>) -> Self {
+        increment_uart_instance(T::info().index);
         Self {
             info: T::info(),
             _tx_dma,
             _phantom: PhantomData,
         }
+    }
+
+    /// Deinitializes a USART Tx instance.
+    pub fn deinit(&self) -> Result<()> {
+        decrement_uart_instance(self.info.index);
+        // This function waits for TX complete, disables TX and RX, and disables the USART clock
+        //while self.info.regs.stat().read().txidle().bit_is_clear() {
+        // When 0, indicates that the transmitter is currently in the process of sending data.
+        //}
+        if self.info.regs.stat().read().txidle().bit_is_clear() {
+            panic!("TX is not idle");
+        }
+
+        // Disable fifo interrupts
+        self.info
+            .regs
+            .fifointenclr()
+            .modify(|_, w| w.txerr().set_bit().txlvl().set_bit());
+
+        // Disable other interrupts
+        self.info.regs.intenclr().write(|w| {
+            w.txidleclr()
+                .set_bit()
+                .framerrclr()
+                .set_bit()
+                .parityerrclr()
+                .set_bit()
+                .aberrclr()
+                .set_bit()
+        });
+
+        // Disable dma requests
+        self.info.regs.fifocfg().modify(|_, w| w.dmatx().clear_bit());
+
+        // Disable FIFOs
+        self.info.regs.fifocfg().modify(|_, w| w.enabletx().disabled());
+
+        if get_uart_instance(self.info.index) == 0 {
+            info!("TESTING: UartTx disabling peripheral");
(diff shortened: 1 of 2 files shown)


akshay-rust:
@jerrysxie @felipebalbi Fixed the review comments received in the deprecated PR https://github.com/OpenDevicePartnership/embassy-imxrt/pull/264  in this PR

src/uart.rs
```diff
@@ -135,12 +135,76 @@ pub type Result<T> = core::result::Result<T, Error>;
...
+
+    /// Deinitializes a USART Tx instance.
+    pub fn deinit(&self) -> Result<()> {
+        decrement_uart_instance(self.info.index);
+        // This function waits for TX complete, disables TX and RX, and disables the USART clock
+        //while self.info.regs.stat().read().txidle().bit_is_clear() {
+        // When 0, indicates that the transmitter is currently in the process of sending data.
+        //}
+        if self.info.regs.stat().read().txidle().bit_is_clear() {
+            panic!("TX is not idle");
```
JamesHuard, line 154: are we sure we want to panic here? Would it be better just to warn and then forceably close the UART peripheral?

src/uart.rs
```diff
@@ -1112,8 +1199,45 @@ pub struct InterruptHandler<T: Instance> {
     _phantom: PhantomData<T>,
 }
 
+/// Uart Errors
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum UartInstances {
```
JamesHuard, line 1205: rather than introduce the statics for manual state tracking, would it be possible to use the type state information in the Uart instance itself? For example, if Tx just disable Tx, if Rx just disable Rx, if it's a combined instance, disable the whole thing? Or is the need to keep the peripheral clocked until both potentials are dropped driving the need for reference counting?

src/uart.rs
jerrysxie: I agreed. We should just stop TX and throw away the data in the FIFO if the user did not maintain ownership of the uart.

src/uart.rs
```diff
@@ -135,12 +135,76 @@ pub type Result<T> = core::result::Result<T, Error>;
...
+
+        if get_uart_instance(self.info.index) == 0 {
+            info!("TESTING: UartTx disabling peripheral");
+            // Note: If Rx, Tx are split, then one of them going out of scope will disable the peripheral.
+            // To prevent that, we need to keep a count of the uart instances (Rx, Tx )being used
+
+            // Disable peripheral
+            self.info.regs.cfg().modify(|_, w| w.enable().disabled());
+
+            // TODO: Cannot call without access to T
```
jerrysxie, line 189: Yeah, I see, we don't have access to FC after `new()`. We can store a pointer to FC register map but it would be a bit hacky.

@felipebalbi @JamesHuard @madeleyneVaca @RobertZ2011 Thoughts?

jerrysxie (changes requested):

src/uart.rs
felipebalbi: I think we can overcome that by preparing a closuring during `new()` and holding it within `struct UartRx` and `struct UartTx`.

jerrysxie: @akshay-rust 

You can try something like this:
```rust
@@ -34,6 +34,7 @@ pub struct I2cMaster<'a, M: Mode> {
     info: Info,
     _phantom: PhantomData<M>,
     dma_ch: Option<dma::channel::Channel<'a>>,
+    clean_up_function: fn(),
 }
 
 impl<'a, M: Mode> I2cMaster<'a, M> {
@@ -94,10 +95,15 @@ impl<'a, M: Mode> I2cMaster<'a, M> {
 
         regs.cfg().write(|w| w.msten().set_bit());
 
+        let function = || {
+            T::info();
+        };
+
         Ok(Self {
             info,
             _phantom: PhantomData,
             dma_ch,
+            clean_up_function: function,
         })
     }
```

See these references for more information on closure/function pointer in Rust:
* https://doc.rust-lang.org/book/ch13-01-closures.html
* https://doc.rust-lang.org/book/ch19-05-advanced-functions-and-closures.html

PR #291, opened 2025-02-04
Enable CTimer based PWM

sukomath:
Addressing review comments from #268


src/timer.rs
```diff
@@ -694,6 +882,152 @@ impl<M: Mode> Drop for CaptureTimer<M> {
...
+    fn set_duty(&mut self, channel: Self::Channel, duty: Self::Duty) {
+        let scaled = duty.as_scaled(self.count_max);
+        let reg = self.info.regs;
+
+        // PWM output is low at the beginning of PWM cycle
+        // PWM output is set to high when timer count reaches match register value
+        // For active high PWM, set match register such that output is high for PWM cycle length*dutycycle
+        channel.channel_start_pwm(self.info.regs);
+
+        unsafe {
```
JamesHuard, line 947: nit: missing //SAFETY:

src/timer.rs
```diff
@@ -694,6 +882,152 @@ impl<M: Mode> Drop for CaptureTimer<M> {
...
+
+impl<'p, T: Instance> CTimerPwmLengthChannel<'p, T> {
+    /// Take the `CTimer` instance supplied and use it as a simple PWM driver. Function returns constructed Pwm instance.
+    pub fn new(length_channel: impl Peripheral<P = T> + 'p, period: MicroSeconds) -> Self {
+        let clock_rate = Hertz(T::info().pwm_get_clock_freq());
+
+        let requested_pwm_rate: Hertz = period.into();
+
+        // we cannot clock faster than the supplied clock rate
+        assert!(period.0 > 0);
```
JamesHuard, line 1015: assert's probably fine here, though was there any consideration to return a Result<Self> instead?

JamesHuard (approved):

src/timer.rs
```diff
@@ -325,6 +450,69 @@ impl Info {
...
+        }
+        freq
+    }
+
+    fn pwm_configure(&self, period: u32) {
+        let reg = self.regs;
+        let len_channel = self.channel;
+
+        // Disable timer for configuration
+        reg.tcr().write(|w| w.cen().clear_bit());
```
jerrysxie, line 489: This would cause a problem if there is an ongoing timer operation on another channel from the same timer.

src/timer.rs
```diff
@@ -325,6 +450,69 @@ impl Info {
...
+
+        // Use length channel to set PWM cycle length
+        //SAFETY: No safety impact as we are writing match register here
+        unsafe {
+            reg.mr(len_channel).write(|w| w.match_().bits(period));
```
jerrysxie, line 494: We can restrict the unsafe block into a smaller scope like this:

```rust
        reg.mr(len_channel).write(|w| 
            //SAFETY: No safety implication for using bits()
            unsafe { w.match_().bits(value) });
```

src/timer.rs
```diff
@@ -325,6 +450,69 @@ impl Info {
...
+    }
+
+    fn pwm_configure(&self, period: u32) {
+        let reg = self.regs;
+        let len_channel = self.channel;
+
+        // Disable timer for configuration
+        reg.tcr().write(|w| w.cen().clear_bit());
+
+        // Use length channel to set PWM cycle length
```
jerrysxie, line 491: Doesn't this update the mr register for the length channel every single time a PWM channel is created? I don't think that is what we want.

src/timer.rs
```diff
@@ -325,6 +450,69 @@ impl Info {
...
+        //SAFETY: No safety impact as we are writing match register here
+        unsafe {
+            reg.mr(len_channel).write(|w| w.match_().bits(period));
+        }
+
+        // Set MRnR bit to enable timer reset for register setting PWM length
+        match TIMER_CHANNELS_ARR[len_channel] {
+            TimerChannelNum::Channel0 => {
+                config_timer_reg_channel_bit!(reg, mcr, mr, 0, r, set_bit);
```
jerrysxie, line 500: We should steer away from using the paste! macro for this if possible, because this is not really saving us anything. I would opt for more verbose and more readable code:

```rust
reg.mcr().modify(|_, w| w.mr0r().set_bit());
```

src/timer.rs
```diff
@@ -812,3 +1146,31 @@ macro_rules! impl_pin {
...
+                self.enable_input_buffer();
+                self.set_input_inverter(Inverter::$invert);
+            }
+        }
+    };
+}
+
+// CTimer Match output pins
+// We can add all the GPIO pins here which can be used as CTimer Match output pins
+impl_pin!(PIO0_20, F4, Enabled);
```
jerrysxie, line 1174: It seems like ctimer pins are statically assigned to each timer as well. We are not enforcing that here. In this current implementation, it is possible for the user to pass in the wrong pins for a timer and it would fail silently.

image

Let's create an issue to track and improve it in a later PR.

jerrysxie (changes requested):

src/timer.rs
sukomath: Added //SAFETY

src/timer.rs
sukomath: Switched to return Result for CTimer and CTimerPwmLengthChannel instantiation 

image

src/timer.rs
sukomath: I removed disabling timer before writing PWM length as match register can be written without disabling timer. For PWM enable, added a check for timer disable before resetting and enabling the timer

image

src/timer.rs
sukomath: Added this change to match register unsafe writes

src/timer.rs
sukomath: Moved writing mr register for PWM period to pwm_configure(). This will be called only when a PWM length channel is created.

src/timer.rs
```diff
@@ -694,6 +882,152 @@ impl<M: Mode> Drop for CaptureTimer<M> {
     }
 }
 
+/// Basic PWM Object, Consumes `CTimer` peripheral hardware instances for match channel and PWM length channel on construction
+pub struct CTimerPwm<'p, T: Instance, U: Instance> {
```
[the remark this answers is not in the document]
sukomath: I have tried to add some changes to reduce typestate. Let me if there is anything that can be optimized

src/timer.rs
sukomath: Added an issue: https://github.com/OpenDevicePartnership/embassy-imxrt/issues/309

src/timer.rs
```diff
@@ -694,6 +882,152 @@ impl<M: Mode> Drop for CaptureTimer<M> {
...
+
+impl<'p, T: Instance, U: Instance> CTimerPwm<'p, T, U> {
+    /// Take the `CTimer` instance supplied and use it as a simple PWM driver. Function returns constructed Pwm instance.
+    pub fn new(
+        match_channel: impl Peripheral<P = T> + 'p,
+        length_channel: &'p CTimerPwmLengthChannel<U>,
+        matchoutput_pin: impl CTimerMatchOutput,
+    ) -> Self {
+        // Assert if length channel and PWM channel does not belong to same CTimer
+        assert!(U::info().module == T::info().module);
```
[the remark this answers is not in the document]
sukomath: Added returning a result.  I couldnt add marker traits as we are passing a reference to the length channel instance with current implementation. 

 let ctimerlenchannel = CTimerPwmLengthChannel::new(p.CTIMER4_COUNT_CHANNEL0, MicroSeconds(10_000)).unwrap();
    let mut ctimer = CTimerPwm::new(p.CTIMER4_COUNT_CHANNEL3, &ctimerlenchannel, p.PIO0_31).unwrap();

Let me know if it is ok to track this in a different issue to track this https://github.com/OpenDevicePartnership/embassy-imxrt/issues/310

examples/rt685s-evk/src/bin/pwm.rs
```diff
@@ -67,34 +76,39 @@ async fn main(_spawner: Spawner) {
     sct0.enable(Channel::Ch6);
 
     loop {
-        info!("PWM: Verify LED is off.");
+        info!("PWM: Verify Blue and Red LEDs are off.");
         let duty = CentiPercent(0, 0);
         sct0.set_duty(Channel::Ch6, duty);
+        ctimer.set_duty(TimerChannelNum::Channel3, duty);
```
jerrysxie, line 82: Did we add the channel parameter because of `embedded_hal_02::Pwm`? This seems to conflict with the design to use peripheral reservation to enforce compile time error.

If we opt for dynamic channel, then we have to add dynamic checking of the channel number? For instance, CTIMER4_COUNT_CHANNEL0 is used as the PWMLengthChannel, but there is nothing preventing TimerChannelNum::Channel3 being passed in here. Don't we also have to switch to a different pin from one in `new()` if a different channel was passed in here?

Instead of needing channel number to be passed in, we should implement `embedded_hal::Pwm` without the additional channel parameter: https://github.com/rust-embedded/embedded-hal/blob/1599b06030aaeeb602b7403aa215bd32affe8ff9/embedded-hal/src/pwm.rs#L82.

The additional channel parameter for these function can all be removed, we know which channel we are operating on from
```
pub struct CTimerPwm<'p> {
    _lifetime: PhantomData<&'p ()>,
    _len: &'p CTimerPwmLengthChannel<'p>, <------------- length channel
    period: MicroSeconds,
    count_max: u32,
    info: Info,
}

struct Info {
    regs: &'static crate::pac::ctimer0::RegisterBlock,
    inputmux: &'static crate::pac::inputmux::RegisterBlock,
    module: usize, <---------------------------- counter module
    channel: usize, <---------------------------- counter channel
}
```

src/timer.rs
```diff
@@ -694,6 +886,165 @@ impl<M: Mode> Drop for CaptureTimer<M> {
     }
 }
 
+/// Basic PWM Object, Consumes `CTimer` peripheral hardware instances for match channel and PWM length channel on construction
+pub struct CTimerPwm<'p> {
+    _lifetime: PhantomData<&'p ()>,
+    _len: &'p CTimerPwmLengthChannel<'p>,
```
jerrysxie, line 892: Can we rename this to _len_channel just to be clear?

src/timer.rs
```diff
@@ -694,6 +886,165 @@ impl<M: Mode> Drop for CaptureTimer<M> {
...
+
+/// Basic PWM Length channel Object, Consumes `CTimer` peripheral hardware instances for match channel and PWM length channel on construction
+pub struct CTimerPwmLengthChannel<'p> {
+    _lifetime: PhantomData<&'p ()>,
+    period: MicroSeconds,
+    count_max: u32,
+    info: Info,
+}
+
+impl embedded_hal_02::Pwm for CTimerPwm<'_> {
```
jerrysxie, line 906: We should implement the embedd_hal::Pwm traits that does not take a channel parameter: https://github.com/rust-embedded/embedded-hal/blob/1599b06030aaeeb602b7403aa215bd32affe8ff9/embedded-hal/src/pwm.rs#L82.

Adding the channel parameter adds runtime checking. Since each CTimerPWM is only associated with a single timer channel, the user can only pass in one valid channel anyway.

jerrysxie (changes requested):

examples/rt685s-evk/src/bin/pwm.rs
sukomath: Channel parameter was added as we are using `embedded_hal_02::Pwm`. 

`embedded-hal-1.0.0` pwm module only has SetDutyCycle trait which configures PWM channels. SetDutyCycle does not support enable/disable channel/set period which is supported by `embedded-hal-0.2.6` Pwm Trait. 

image

image

Enable/disable channel could be useful when we want to temporarily halt and resume PWM output. I was mostly following the design from SCTimer PWM. @JamesHuard Is there a reason for SCTimer to use `embedded-hal-0.2.6` Pwm instead of `embedded-hal-1.0.0` pwm module?

JamesHuard: I think I mainly used the 0.2.6 as it had the disable and period APIs, which seemed like a broader stroke abstraction than the 1.0 version at the time and was an easier map to the SCTimer hardware. There isn't necessarily a reason we couldn't use 1.0

src/timer.rs
jerrysxie: or maybe even `_period_channel` because it is just the channel to provide the period of the pulse.

sukomath: Renaming to period_channel

examples/rt685s-evk/src/bin/pwm.rs
sukomath: As we have precision concerns for duty cycle calculation when using embedded_hal_1 PWM APIs, added unused variable changes for channel parameter

src/timer.rs
```diff
@@ -1,28 +1,155 @@
...
+                    reg.mcr().modify(|_, w| w.mr0r().clear_bit());
+                    reg.mcr().modify(|_, w| w.mr0s().clear_bit());
+                    reg.mcr().modify(|_, w| w.mr0i().clear_bit());
+
+                    reg.emr().modify(|_, w| w.em0().clear_bit());
+                    reg.emr().modify(|_, w| w.emc0().clear());
+
+                    reg.ir().modify(|_, w| w.mr0int().clear_bit());
+                } else {
+                    reg.pwmc().modify(|_, w| w.pwmen0().clear_bit());
```
jerrysxie, line 73: We can use the enumerate variants: `match_()` and `pwm()` to set this field.

src/timer.rs
```diff
@@ -1,28 +1,155 @@
...
+    fn channel_start_pwm(&self, reg: &'static crate::pac::ctimer0::RegisterBlock) {
+        use TimerChannelNum::{Channel0, Channel1, Channel2, Channel3};
+
+        // To start PWM output on a channel:
+        // 1. Set PWM enable bit in PWM control register
+        // 2. Clear interrupt flag
+        // 3. Enable interrupt generation when match register matches the value in TC
+        match self {
+            Channel0 => {
+                reg.pwmc().modify(|_, w| w.pwmen0().set_bit());
```
jerrysxie, line 130: We can use the enumerate variants: `match_()` and `pwm()` to set this field.

src/timer.rs
```diff
@@ -1,28 +1,155 @@
...
+            Channel1 => {
+                if isenable {
+                    reg.mcr().modify(|_, w| w.mr1r().clear_bit());
+                    reg.mcr().modify(|_, w| w.mr1s().clear_bit());
+                    reg.mcr().modify(|_, w| w.mr1i().clear_bit());
+
+                    reg.emr().modify(|_, w| w.em1().clear_bit());
+                    reg.emr().modify(|_, w| w.emc1().clear());
+
+                    reg.ir().modify(|_, w| w.mr1int().clear_bit());
```
jerrysxie, line 85: This seems to be a write 1 to clear register, which writing zero has no effect. We need to set_bit() to clear it.

May you create an issue so we patch the PAC to fix this?

src/timer.rs
```diff
@@ -1,28 +1,155 @@
...
+    /// PWM output with min precision cannot enabled with provided period and clock rate
+    PwmPrecisionNotSupported,
+
+    /// Pwm length channel and output channel does not belong to same CTimer
+    PwmChannelMismatch,
+}
+
+impl TimerChannelNum {
+    fn channel_en_pwm(&self, reg: &'static crate::pac::ctimer0::RegisterBlock, isenable: bool) {
+        use TimerChannelNum::{Channel0, Channel1, Channel2, Channel3};
```
jerrysxie, line 51: This is probably not needed.

src/timer.rs
```diff
@@ -1,28 +1,155 @@
...
+                    reg.ir().modify(|_, w| w.mr3int().clear_bit());
+                } else {
+                    reg.pwmc().modify(|_, w| w.pwmen3().clear_bit());
+                }
+            }
+        }
+    }
+
+    fn channel_start_pwm(&self, reg: &'static crate::pac::ctimer0::RegisterBlock) {
+        use TimerChannelNum::{Channel0, Channel1, Channel2, Channel3};
```
jerrysxie, line 122: This line is probably not needed.

src/timer.rs
```diff
@@ -694,6 +877,165 @@ impl<M: Mode> Drop for CaptureTimer<M> {
...
+        let clock_rate = Hertz(channel_info.pwm_get_clock_freq());
+
+        let requested_pwm_rate: Hertz = period.into();
+
+        // we cannot clock faster than the supplied clock rate
+        if period.0 == 0 {
+            return Err(Error::InvalidPwmPeriod);
+        }
+        // assure precision is possible (10_000 ticks within PWM minimum)
+        if requested_pwm_rate.0 > clock_rate.0 / 10_000 {
```
jerrysxie, line 1020: This is to ensure we at least have 10,000 tick per period, so ensure the precise of the period? The comment's wording can be updated to make this more clear.

src/timer.rs
```diff
@@ -694,6 +877,165 @@ impl<M: Mode> Drop for CaptureTimer<M> {
...
+            return Err(Error::PwmPrecisionNotSupported);
+        }
+
+        // 10_000 clocks per period to achieve the desired precision.
+        let factor = clock_rate.0 / requested_pwm_rate.0;
```
jerrysxie, line 1025: This factor seems like period in ticks? Renaming it would make this more clear.

src/timer.rs
```diff
@@ -694,6 +877,165 @@ impl<M: Mode> Drop for CaptureTimer<M> {
...
+    }
+
+    fn set_duty(&mut self, _: (), duty: Self::Duty) {
+        let scaled = duty.as_scaled(self.count_max);
+        let reg = self.info.regs;
+
+        // PWM output is low at the beginning of PWM cycle
+        // PWM output is set to high when timer count reaches match register value
+        // For active high PWM, set match register such that output is high for PWM cycle length*dutycycle
+        TIMER_CHANNELS_ARR[self.info.channel].channel_start_pwm(self.info.regs);
```
jerrysxie, line 942: Assuming the PWM is already enabled, then this call does nothing, right? What is the default duty cycle when 'enable()' is called withou setting the duty cycle?

src/timer.rs
```diff
@@ -694,6 +877,165 @@ impl<M: Mode> Drop for CaptureTimer<M> {
...
+        reg.mr(self.info.channel).write(|w|
+            //SAFETY: No safety impact as we are writing match register here
+            unsafe { w.match_().bits(self.count_max - scaled)});
```
jerrysxie, line 945: Can we add some comments on why this will not glitch if we change it on a live PWM?

image

Reading the documentation, it seems like if we low, then we will go high when we reach the new match value. If we are high already, then we will stay high and update the duty cycle on the next period. Is that the behavior we see in testing?

src/timer.rs
```diff
@@ -694,6 +877,165 @@ impl<M: Mode> Drop for CaptureTimer<M> {
...
+
+        // update scale factor
+        self.count_max = clock_rate.0 / requested_pwm_rate.0;
+
+        // Set period through match register
+        self.info.pwm_configure(self.count_max);
+
+        // update duty cycle match registers according to new scale factor
+        (0..TIMER_CHANNELS_ARR.len()).for_each(|i| {
+            self.set_duty((), duty_cycles[i]);
```
jerrysxie, line 971: This is a very dangerous call, even though our ctimer are separate channels. Updating the period of one channel affects all the other channels on the same timer.

Let's put a comment to indicate that this call has subtle consequence affecting all channels and create an issue to fix up this design later.

What happens when we calling this on a live PWM, all 3 channels will glitch together?

src/timer.rs
jerrysxie: Why is 10,000 tick the desired resolution? That is very limiting.

jerrysxie (changes requested):

src/timer.rs
sukomath: Using embedded_hal_02 as per discussion for duty cycle config precision. Added unused parameter for channel.

src/timer.rs
```diff
@@ -1,28 +1,155 @@
...
+    InvalidPwmPeriod,
+
+    /// PWM output with min precision cannot enabled with provided period and clock rate
+    PwmPrecisionNotSupported,
+
+    /// Pwm length channel and output channel does not belong to same CTimer
+    PwmChannelMismatch,
+}
+
+impl TimerChannelNum {
```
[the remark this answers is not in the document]
sukomath: Moving everything in 'impl TimerChannelNum' to CTimerPwm.

src/timer.rs
```diff
@@ -1,28 +1,155 @@
...
+    fn channel_en_pwm(&self, reg: &'static crate::pac::ctimer0::RegisterBlock, isenable: bool) {
```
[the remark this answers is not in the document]
sukomath: Adding two functions channel_enable_pwm() and channel_disable_pwm()

src/timer.rs
sukomath: Switched to using variants

` reg.pwmc().modify(|_, w| w.pwmen0().match_());`
` reg.pwmc().modify(|_, w| w.pwmen0().pwm());`

src/timer.rs
sukomath: Changed the implementation - previously enable() did not set duty cycle. It was required to call set_duty() to set duty cycle and start PWM. In the current implementation, enable() sets a default duty cycle of 0 and starts the PWM. 
set_duty() will only update duty cycle, it will not start PWM.

src/timer.rs
sukomath: Changing this to 500 ticks to support required fan PWM frequency. I will add a task in ADO to verify PWM precision requirements.

src/timer.rs
sukomath: Renamed to period_clock_ticks

src/timer.rs
sukomath: From testing set duty cycle on live PWM,

- Initial config, PWM output configured on two channels of same CTimer. PWM output 1 has duty cycle 75% and PWM output 2 has duty cycle 25%. PWM period = 10000 microseconds

image

- At marker 1, setduty() is called for PWM1 to set duty cycle to 25% and for PWM2 to 75%
- At marker 1, PWM1 output is high and continues to be high till end of current PWM cycle. New duty cycle seen in PWM output 1 from marker2(end of PWM cycle)
- At marker 1, PWM2 output is low and continues to be low till end of current PWM cycle. New duty cycle seen in PWM output 2 from marker2(end of PWM cycle)

Set duty cycle will not disrupt a live PWM immediately. New duty cycle seems to be configured after the current PWM cycle ends. However, changing period on live PWM using setperiod() has different behavior. 

Set period test on live PWM

- Initial config, PWM output configured on two channels of same CTimer. PWM output 1 has duty cycle 75% and PWM output 2 has duty cycle 25%. PWM period = 10000 microseconds

image

- At marker 1, setperiod() is called to change PWM period to 20000 microseconds. This will write the new PWM length to period channel match register. Existing duty cycles are adjusted to match new PWM length. 
- PWM 1 is high at marker 1 and continues to be high till new PWM length. New duty cycle is seen immediately after configuration.
- PWM 2 is low at marker 1 and continues to be low till match value corresponds to new duty cycle. Upon match, PWM2 goes high and stays high till new PWM length. New duty cycle is seen immediately after configuration.

src/timer.rs
sukomath: Set period test on live PWM

- Initial config, PWM output configured on two channels of same CTimer. PWM output 1 has duty cycle 75% and PWM output 2 has duty cycle 25%. PWM period = 10000 microseconds

image

- At marker 1, setperiod() is called to change PWM period to 20000 microseconds. This will write the new PWM length to period channel match register. Existing duty cycles are adjusted to match new PWM length. 
- PWM 1 is high at marker 1 and continues to be high till new PWM length. New duty cycle is seen immediately after configuration.
- PWM 2 is low at marker 1 and continues to be low till match value corresponds to new duty cycle. Upon match, PWM2 goes high and stays high till new PWM length. New duty cycle is seen immediately after configuration.

src/timer.rs
sukomath: I added a comment in set duty about the behavior on an already running PWM. Let me know if need to do any further analysis

jerrysxie: Thanks for doing the testing, it seems like in the case where the new match is lower than current count of ticks within the period, we would glitch with an out of spec pulse. There are some tricks we can used to deal with this, but just adding comment and documenting it would be good enough for now.

src/timer.rs
jerrysxie: In set_period(), we update the periodic and then the duty cycle. So technically it is possible to run into the same out of spec pulse when period is updated.

The biggest issue right now is that updating the period for one channel affects the other 2 channels. Let's document it and create an issue to figure out a better way to deal with this in the future.

src/timer.rs
sukomath: Can we use do_nothing() variant to write 0 to this register? We want to set emc bit to 2 so that match output will go high when there is match between TC and match register. I don't see any impact for this bit on match output. Adding an issue to track this: https://github.com/OpenDevicePartnership/embassy-imxrt/issues/334

jerrysxie: Sorry, I was referring to the ir register, not the emr register:
image

src/timer.rs
```diff
@@ -694,6 +774,250 @@ impl<M: Mode> Drop for CaptureTimer<M> {
...
+
+        CentiPercent::from_scaled(self.count_max - scaled, self.count_max)
+    }
+
+    fn get_max_duty(&self) -> Self::Duty {
+        CentiPercent::MAX
+    }
+
+    fn set_duty(&mut self, _: (), duty: Self::Duty) {
+        // PWM output is not disrupted by set_duty. New duty cycle will be reflected after current PWM cycle ends.
```
jerrysxie, line 909: From your testing: https://github.com/OpenDevicePartnership/embassy-imxrt/pull/291#discussion_r1974662747, it does produce an out of spec pulse when the new duty cycle set a match value that is lower than the current tick count within the period, right? The pulse would stay low, the count for the period resets, and it would only go high with the new duty cycle. So that first pulse after setting would be out of spec.

We should document in the comment that at most one out of spec pulse after update it is possible here and same for `set_period()`. I am not sure if that will be problem for the fan we are using. We might need to address it if this cause weird fan behavior.

jerrysxie (changes requested):

src/timer.rs
sukomath: Updated ir().modify to set_bit instead of clear_bit. Also added issue in pac to track this: https://github.com/OpenDevicePartnership/mimxrt633s-pac/issues/20

src/timer.rs
sukomath: Added comment in set_period() to describe the impact of setting period on a live PWM. Added issue: https://github.com/OpenDevicePartnership/embassy-imxrt/issues/335

src/timer.rs
sukomath: Added comment in set_duty to describe impact on already running PWM

jerrysxie (approved):
Great work! 👍

PR #292, opened 2025-02-04
Handle i2c master read and write of more than 1024 bytes

### LINKED ISSUE #255: I2C transaction needs to handle DMA transfer size large than 1024
Working on USB PD firmware update, we encountered the problem where DMA transfer for i2c cannot exceed 1024 bytes. That makes sense as DMA transfer size maximum is 1024. We should handle this case better.

There are two approaches:

- START -> write/read a block of 1024 bytes -> RESTART-> write/read a block 1024 bytes -> ... -> STOP
  This is can be achieved with `transaction()` manually and break up the writes in `write()`/reads in `read()` like this
  ```rust
      async fn write(&mut self, address: u8, write: &[u8]) -> Result<()> {
        let num_block = write.len() / 1024;
        for i in 0..num_block {
            info!("writing {}..{}", i * 1024, (i + 1) * 1024);
            self.write_no_stop(address, &write[(i * 1024)..((i + 1) * 1024)])
                .await?;
        }
        info!("writing {}..{}", num_block * 1024, write.len());
        self.write_no_stop(address, &write[(num_block * 1024)..]).await?;

        self.stop().await
    }
  ```

  Another possible approach is to limit individual calls for `write()` and `read()` to be lower than 1024 bytes and returning an error if it is greater. For writes greater than 1024 bytes, force the HAL user to call `transaction()`

- START -> write/read a block of 1024 bytes -> write/read a block 1024 bytes -> ... -> STOP
  This will require more work inside the i2c `write()` + `read()` to do the DMA transfer in a loop. This will be a little bit tricky due to the inherent race conditions we had to work around in the NXP i2c peripheral.

akshat2112:
Handling the case of more than 1024 bytes in i2c read/write by dividing it into chunks of 1024
This PR consists of master side changes. Slave changes to be made in a separate PR

Closes #255


src/i2c/master.rs
```diff
@@ -617,18 +617,26 @@ impl<M: Mode> embedded_hal_1::i2c::ErrorType for I2cMaster<'_, M> {
 // implement generic i2c interface for peripheral master type
 impl embedded_hal_1::i2c::I2c for I2cMaster<'_, Blocking> {
     fn read(&mut self, address: u8, read: &mut [u8]) -> Result<()> {
-        self.read_no_stop(address, read)?;
+        for chunk in read.chunks_mut(1024) {
```
jerrysxie, line 620: Blocking does not use DMA, so we don't need to chunk data into 1024-byte chunks.

src/i2c/master.rs
```diff
@@ -656,18 +664,26 @@ impl embedded_hal_1::i2c::I2c for I2cMaster<'_, Blocking> {
...
-        self.read_no_stop(address, read).await?;
+        for chunk in read.chunks_mut(1024) {
+            self.read_no_stop(address, chunk).await?;
+        }
         self.stop().await
     }
 
     async fn write(&mut self, address: u8, write: &[u8]) -> Result<()> {
-        self.write_no_stop(address, write).await?;
+        for chunk in write.chunks(1024) {
```
jerrysxie, line 674: So we opted for START - write - RESTART -write - ... ... - STOP. That is the most efficient way since we don't need to do STOP between each write.

Can you sync up with @RobertZ2011 to test this with PD firmware update? Because that is one of the few cases that will exercise this.

jerrysxie (changes requested):
What about i2c slave? We should also add an extra loopback example where we are sending and receiving more than 1024 bytes of unique data so we can make sure the data transferred is correct?

For first 1024 bytes, we can send all 1s
For second 1024 bytes, we can send all 2s
...

src/i2c/master.rs
RobertZ2011: The FW update code now limits its writes to 255 bytes to make working with CFU easier.

felipebalbi (changes requested):
It looks like you need to rebase on odp main. Not merge it.

PR #305, opened 2025-02-11
Add i2c master 10-bit address support

### LINKED ISSUE #71: I2C - 10 bit address support
currently only supporting 7-bit (default) addressing

### LINKED ISSUE #250: I2C 10 bit address support - master

akshat2112:
Adding 10-bit addressing support for I2C master.
New PR because of branch remapping. Old PR at #263 
Closes #250

### DIFF
--- a/src/i2c/master.rs
+++ b/src/i2c/master.rs
@@ -9,7 +9,7 @@ use embassy_hal_internal::into_ref;
 
 use super::{
     Async, Blocking, Error, Info, Instance, InterruptHandler, MasterDma, Mode, Result, SclPin, SdaPin, TransferError,
-    I2C_WAKERS,
+    I2C_WAKERS, TEN_BIT_PREFIX,
 };
 use crate::interrupt::typelevel::Interrupt;
 use crate::{dma, interrupt, Peripheral};
@@ -133,7 +133,19 @@ impl<'a> I2cMaster<'a, Blocking> {
         Ok(this)
     }
 
-    fn start(&mut self, address: u8, is_read: bool) -> Result<()> {
+    fn start(&mut self, address: u16, is_read: bool) -> Result<()> {
+        // check if the address is 10-bit
+        let is_10bit = address > 0x7F;
+
+        // start with the correct address
+        if is_10bit {
+            self.start_10bit(address, is_read)
+        } else {
+            self.start_7bit(address as u8, is_read)
+        }
+    }
+
+    fn start_7bit(&mut self, address: u8, is_read: bool) -> Result<()> {
         let i2cregs = self.info.regs;
 
         self.poll_ready()?;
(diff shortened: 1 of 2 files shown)


src/i2c/master.rs
```diff
@@ -169,7 +181,64 @@ impl<'a> I2cMaster<'a, Blocking> {
...
+
+        // If this is a read operation, send a repeated start with the read bit set
+        if is_read {
+            let addr_high_read = addr_high | 0b1;
+            i2cregs.mstdat().write(|w| unsafe { w.data().bits(addr_high_read) });
+            i2cregs.mstctl().write(|w| w.mststart().set_bit());
+
+            self.poll_ready()?;
+            self.check_for_bus_errors()?;
+        }
```
felipebalbi, line 228: This is the only real difference between 7bit and 10bit. Do we need all the other duplication? How about combining them? Check for address errors earlier and bail out, then right at this point, you check if the high byte has anything, if it does, send the other half of the address and continue. The rest of this function is also the same as 7bit.

felipebalbi (changes requested):

src/i2c/slave.rs
```diff
@@ -111,11 +154,29 @@ impl<'a, M: Mode> I2cSlave<'a, M> {
...
+                // address 0 match = addr, per UM11147 24.3.2.1
+                i2c.slvadr(0).modify(|_, w|
+                        // note: shift is omitted as performed via w.slvadr() 
+                        // SAFETY: unsafe only required due to use of unnamed "bits" field
+                        unsafe{w.slvadr().bits(addr)}.sadisable().enabled());
+            }
+            Address::TenBit(_addr) => {
+                // 10 bit address have first byte between 0b1111_0000 and 0b1111_0110 as per spec
+                // using slvadr(0) and slvqual0 to check, per UM11147 24.6.15
+                i2c.slvqual0().write(|w| w.qualmode0().extend());
```
jerrysxie, line 168: The i2c slave part will still not work like we expected. We are matching the first byte here, but we are not matching the second byte of the address. It will require changes to both read and write:
image

For write, we have to check the next byte to make sure the lower byte match:
image

For read, we have to check the second byte, get a restart, then match the first upper byte again:
image

Both will require changes to how i2c slaves handles writes and reads.

At this point, I believe the i2c master is at a good spot, so we should separate out the slave portion into its own PR so that the master part can go in. Because the slave portion will require more changes.

jerrysxie (changes requested):

jerrysxie (approved):

src/i2c/master.rs
akshat2112: Introduced i2c address trait to handle both 7 bit and 10 bit addresses. This enables us to avoid duplication of read/write functions for each address type.
With this change, the start function is the only one with some level of duplication but since that's the main difference in handling 10 bit addresses, I believe it makes sense to have separate start for both types of addresses.

RobertZ2011 (approved):

src/i2c/master.rs
jerrysxie: I am not a fan of the new extra trait `I2cAddress`, but I see that it does reduce some code duplication. I believe @felipebalbi is commenting about something different though.

jerrysxie (approved):

src/i2c/master.rs
akshat2112: Yes, @felipebalbi is talking about duplication in the start function. I see the start function as being completely different for 7 bit and 10 bit addresses, so opting to keep them separate.

src/i2c/master.rs
```diff
@@ -619,34 +836,46 @@ impl<M: Mode> embedded_hal_1::i2c::ErrorType for I2cMaster<'_, M> {
...
+    fn to_u16(self) -> u16 {
+        self as u16
+    }
+}
+
+impl I2cAddress for u16 {
+    fn to_u16(self) -> u16 {
+        self
+    }
+}
```
felipebalbi, line 849: I think this is unnecessary. You know the types can only be `u16` or `u8`. In both cases there is a `u16::from()` implementation. Just unconditionally call `addr.into()` and you should be good, no?

src/i2c/master.rs
felipebalbi: Yeah, I still think there's a way to combine them and not look so weird, but that's okay. I was, indeed, commenting about the duplication in the start function.

PR #311, opened 2025-02-13
Add i2c slave 10-bit address support

### LINKED ISSUE #251: I2C 10 bit address support - slave

akshat2112:
This PR enables support for 10 bit addressing for i2c slave.
Master changes part of #305 

Closes #251


felipebalbi (approved):

examples/rt685s-evk/src/bin/i2c-loopback-async-10bit.rs
```diff
@@ -0,0 +1,125 @@
...
+
+        if i == 30 {
+            info!("i2c 10 bit loopback test end");
+            break;
+        }
+    }
+}
+
+#[embassy_executor::main]
+async fn main(spawner: Spawner) {
```
jerrysxie, line 115: We should validate the data in this example to make sure that the second byte of the address did not get slipped in as part of the data.

src/i2c/slave.rs
```diff
@@ -8,42 +8,100 @@ use embassy_hal_internal::{into_ref, Peripheral};
...
+}
+
+#[derive(Copy, Clone, Debug)]
+struct TenBitAddressInfo {
+    first_byte: u8,
+    second_byte: u8,
+}
+
+impl TenBitAddressInfo {
+    fn new(first_byte: u8, second_byte: u8) -> Self {
```
jerrysxie, line 100: Does this make more sense to take in a u16 instead? Then the logic to convert from u16 addr to 2 separate bytes can be encapsulate in new(). Internally, storing it as 2 separate u8s is fine.

src/i2c/slave.rs
```diff
@@ -336,6 +419,23 @@ impl I2cSlave<'_, Async> {
         // Poll for HW to transitioning from addressed to receive/transmit
         self.poll_sw_action().await;
 
+        if let Some(info) = self.ten_bit_info {
```
jerrysxie, line 439: Can we rename info to ten_bit_addre? Since there is already an info field with self, this is a bit confusing.

src/i2c/slave.rs
```diff
@@ -336,6 +419,23 @@ impl I2cSlave<'_, Async> {
...
+            // For 10 bit address, the first byte received is the second byte of the address
```
jerrysxie, line 440: The behavior below is a bit contradicting.

- For first start, it says that we read the second byte of address in from slvdat. <--- I would expect the first read would get first address byte just like the second restart, can we double check that? And can we do data verification in the example to make sure that we actually did 10-bits correctly?

- For the second restart, now it says that read the first byte of data from slvdat.

src/i2c/slave.rs
```diff
@@ -336,6 +419,23 @@ impl I2cSlave<'_, Async> {
...
+            if i2c.slvdat().read().data().bits() == info.second_byte {
+                i2c.slvctl().write(|w| w.slvcontinue().continue_());
+                self.poll_sw_action().await;
+            }
+
+            // Check for a restart
+            if i2c.stat().read().slvstate().is_slave_address() {
```
jerrysxie, line 447: If it is a restart, then we know it is a read, right? If it is not a restart, then we know it is a write. Can we add some error checking to this code? If the address does not match after the restart, that would be an error case, etc.

Consider the else case and don't let it fall through if it is an error case.

src/i2c/slave.rs
```diff
@@ -285,6 +285,23 @@ impl I2cSlave<'_, Blocking> {
         //Block until we know it is read or write
         self.poll()?;
 
+        if let Some(info) = self.ten_bit_info {
+            // For 10 bit address, the first byte received is the second byte of the address
+            if i2c.slvdat().read().data().bits() == info.second_byte {
```
jerrysxie, line 290: Same comment as the async case, please double check that we getting the second byte in this read.

Please consider possible error conditions.

examples/rt685s-evk/src/bin/i2c-loopback-async-10bit.rs
```diff
@@ -0,0 +1,125 @@
+#![no_std]
```
jerrysxie, line 1: Nit: The example might be better served if it is in a separate commit.

jerrysxie (changes requested):

src/i2c/slave.rs
akshat2112: Makes sense. Done

src/i2c/slave.rs
akshat2112: Done

src/i2c/slave.rs
akshat2112: The main difference between the first address byte matching in first start and second restart is that it's matched by the hardware in first start and software in second restart.
The key field to note is slvcontinue. Setting the continue bit clears the slvdat register and pulls in the next byte.

- For first start, by this point in the code we have already matched the first byte of address (using slvadr register) and set continue, which gets us the first data byte in slvdat register, which is in fact the second address byte for 10 bit addressing.
- Next we immediately check for a restart. If it's a restart, we'll have the first address byte with read bit set in slvdat. We validate it in software and set continue bit, which now triggers the read operation.

This behavior was verified with print messages as well as logic analyzer. I'll add a check in the example to validate further as suggested.

examples/rt685s-evk/src/bin/i2c-loopback-async-10bit.rs
akshat2112: Moved to a separate commit

examples/rt685s-evk/src/bin/i2c-loopback-async-10bit.rs
akshat2112: Added `assert!` statements to validate all read and write buffers

src/i2c/slave.rs
akshat2112: Added else case error handling

src/i2c/slave.rs
akshat2112: Added else case error handling

akshat2112:
Commit cc485b79fab0d7c7316d1457a5712db391b4111e is cherry picked from PR #305 to enable compilation and testing

src/i2c/slave.rs
jerrysxie: This seems a little bit off:
image

Since we are programming SLVQUAL0 + ALVADDR0 to match a range of address in the upper byte, don't we have to manually check the first address byte to make sure the upper 2 bits of the address actually match? If we don't match right now, wouldn't the top 2 bits of the address be off when matching 10-bit writes?

jerrysxie: I think I see what the problem is for 10-bit

In
```
        if i2c.stat().read().slvstate().is_slave_address() {
            i2c.slvctl().write(|w| w.slvcontinue().continue_());
        } else {
            // If we are not addressed here, then we have issues.
            return Err(TransferError::OtherBusError.into());
        }
```

We are hitting continue before we are actually checking the first byte so later on we are only getting the second byte in the initial slvdat read.

Same problem exists for both async and blocking implementation.

akshat2112: I see what the confusion was. On further analysis, I don't think we need to use the SLVQUAL0 method to capture the range of 10-bit addresses in this implementation. We can extract the upper 2 bits of the address, append them to the 10-bit prefix `0b11110` and write to SLVADDR0. This way the existing hardware match and continue is valid for both 7 and 10-bit addresses.

RobertZ2011 (approved):

src/i2c/slave.rs
akshat2112: Updated

src/i2c/slave.rs
```diff
@@ -414,19 +434,25 @@ impl I2cSlave<'_, Async> {
         // Poll for HW to transitioning from addressed to receive/transmit
```
jerrysxie, line 434: This is also changing async, contradictory to the commit message.

src/i2c/slave.rs
```diff
@@ -280,6 +277,29 @@ impl I2cSlave<'_, Blocking> {
...
         self.poll()?;
 
+        if let Some(ten_bit_address) = self.ten_bit_info {
+            // For 10 bit address, the first byte received is the second byte of the address
+            if i2c.slvdat().read().data().bits() == ten_bit_address.second_byte {
+                i2c.slvctl().write(|w| w.slvcontinue().continue_());
+                self.poll()?;
+            } else {
+                // If the second byte of the 10 bit address is not received, then we have issues.
+                return Err(TransferError::OtherBusError.into());
```
jerrysxie, line 287: This it not an bus error per se. If the master sends down an address that does not match our address, we should NACK it.

src/i2c/slave.rs
```diff
@@ -280,6 +277,29 @@ impl I2cSlave<'_, Blocking> {
...
+
+            // Check for a restart
+            if !i2c.stat().read().slvdesel().is_deselected() && i2c.stat().read().slvstate().is_slave_address() {
+                // Check if first byte of 10 bit address is received again with read bit set
+                if i2c.slvdat().read().data().bits() == ten_bit_address.first_byte | 1 {
+                    i2c.slvctl().write(|w| w.slvcontinue().continue_());
+                    self.poll()?;
+                } else {
+                    // If the first byte of the 10 bit address is not received again, then we have issues.
+                    return Err(TransferError::OtherBusError.into());
```
jerrysxie, line 298: If this address byte does not match, the slave should send a NACK as well. It means that the master is doing something wrong but we should be able to send a NACK and be able to recover on the next transaction.

It would be a good idea to manually test this out by simulating this test case in our loopback test.

src/i2c/slave.rs
```diff
@@ -414,19 +434,25 @@ impl I2cSlave<'_, Async> {
...
-                if i2c.slvdat().read().data().bits() == info.first_byte | 1 {
+                if i2c.slvdat().read().data().bits() == ten_bit_address.first_byte | 1 {
                     i2c.slvctl().write(|w| w.slvcontinue().continue_());
                     self.poll_sw_action().await;
+                } else {
+                    // If the first byte of the 10 bit address is not received again, then we have issues.
+                    return Err(TransferError::OtherBusError.into());
                 }
             }
         }
```
jerrysxie, line 458: We should do some sanity checking for the slave receive and slave transmit check below, because for 10-bit we should already know whether it is a write or a read by this point.

src/i2c/slave.rs
```diff
@@ -414,19 +434,25 @@ impl I2cSlave<'_, Async> {
...
-        if let Some(info) = self.ten_bit_info {
+        if let Some(ten_bit_address) = self.ten_bit_info {
             // For 10 bit address, the first byte received is the second byte of the address
-            if i2c.slvdat().read().data().bits() == info.second_byte {
+            if i2c.slvdat().read().data().bits() == ten_bit_address.second_byte {
                 i2c.slvctl().write(|w| w.slvcontinue().continue_());
                 self.poll_sw_action().await;
+            } else {
+                // If the second byte of the 10 bit address is not received, then we have issues.
+                return Err(TransferError::OtherBusError.into());
```
jerrysxie, line 444: Same comment as the blocking case.

src/i2c/slave.rs
```diff
@@ -414,19 +434,25 @@ impl I2cSlave<'_, Async> {
...
             // Check for a restart
-            if i2c.stat().read().slvstate().is_slave_address() {
+            if !i2c.stat().read().slvdesel().is_deselected() && i2c.stat().read().slvstate().is_slave_address() {
                 // Check if first byte of 10 bit address is received again with read bit set
-                if i2c.slvdat().read().data().bits() == info.first_byte | 1 {
+                if i2c.slvdat().read().data().bits() == ten_bit_address.first_byte | 1 {
                     i2c.slvctl().write(|w| w.slvcontinue().continue_());
                     self.poll_sw_action().await;
+                } else {
+                    // If the first byte of the 10 bit address is not received again, then we have issues.
```
jerrysxie, line 454: Same comment as the blocking case.

src/i2c/slave.rs
```diff
@@ -280,6 +277,29 @@ impl I2cSlave<'_, Blocking> {
...
+                    return Err(TransferError::OtherBusError.into());
+                }
+            }
+        }
+
         // We are already deselected, so it must be an 0 byte write transaction
         if i2c.stat().read().slvdesel().is_deselected() {
             // Clear the deselected bit
```
jerrysxie, line 305: We should do some sanity checking for the slave receive and slave transmit check below, because for 10-bit we should already know whether it is a write or a read by this point.

jerrysxie (changes requested):

src/i2c/master.rs
```diff
@@ -397,45 +397,7 @@ impl<'a> I2cMaster<'a, Async> {
 
         i2cregs.mstctl().write(|w| w.mststart().set_bit());
 
-        let res = self
```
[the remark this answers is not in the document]
akshat2112: I'll move this to the master changes PR #305

src/i2c/slave.rs
akshat2112: Fixed the commits

src/i2c/slave.rs
akshat2112: Added address NACK and tested using examples

src/i2c/slave.rs
akshat2112: Added checks for expected transmit/receive state for 10 bit addressed slave

src/i2c/slave.rs
akshat2112: Updated

src/i2c/slave.rs
akshat2112: Updated

src/i2c/slave.rs
akshat2112: Updated

src/i2c/slave.rs
akshat2112: Added address NACK error

src/i2c/slave.rs
```diff
@@ -336,6 +411,45 @@ impl I2cSlave<'_, Async> {
...
+        if let Some(ten_bit_address) = self.ten_bit_info {
+            // For 10 bit address, the first byte received is the second byte of the address
+            if i2c.slvdat().read().data().bits() == ten_bit_address.second_byte {
+                i2c.slvctl().write(|w| w.slvcontinue().continue_());
+                self.poll_sw_action().await;
+            } else {
+                // If the second byte of the 10 bit address is not received, then nack the address.
+                i2c.slvctl().write(|w| w.slvnack().nack());
+                self.poll_sw_action().await;
+                return Err(TransferError::AddressNack.into());
```
jerrysxie, line 462: Address NACK is actually a normal part of I2C slave operation, because there are only 2 wires and the master is broadcasting. So if there is multiple slaves on the bus, address NACK from the other slaves is expected.

So for all the address NACK case, we don't have to return an error. We will need to return from here with no error and go into `listen()` again, so maybe we should just return `Command::Probe` and exit.

This applies to the address NACK below and the blocking implementation as well.

src/i2c/slave.rs
```diff
@@ -336,6 +411,45 @@ impl I2cSlave<'_, Async> {
...
+            }
+
+            // Check this is not a 0 byte write transaction
```
jerrysxie, line 465: Checking is_deselected() is checking whether the master has sent a STOP. This comment is a bit misleading, Instead of checking slvdesel(), we can check slvsel instead() to do away with the negative logic.

The multi-layer of nested if statement below is tough to read, but I can't think of a better way to express these checks either.

jerrysxie (changes requested):

src/i2c/slave.rs
akshat2112: Updated.
I did try implementing something other than nested ifs initially but couldn't think of anything cleaner. In the current implementation, we can expect a success case to simply go through the if statements. The else cases are error handling.

jerrysxie (approved):

PR #314, opened 2025-02-14
add support for buffered uart

huseinmsft:
Implement a circular buffer for UART DMA for RX mode to allow continuous reading or data on UART RX

### DIFF
--- a/src/dma/channel.rs
+++ b/src/dma/channel.rs
@@ -124,24 +124,6 @@ impl<'d> Channel<'d> {
         let xfercount = (mem_len / xferwidth) - 1;
         let channel = self.info.ch_num;
 
-        // Configure the channel descriptor
-        // NOTE: the DMA controller expects the memory buffer end address but peripheral address is actual
-        // SAFETY: unsafe due to use of a mutable static (DESCRIPTORS.list)
-        unsafe {
-            DESCRIPTORS.list[channel].reserved = 0;
-            if dir == Direction::MemoryToPeripheral {
-                DESCRIPTORS.list[channel].dst_data_end_addr = dstbase as u32;
-            } else {
-                DESCRIPTORS.list[channel].dst_data_end_addr = dstbase as u32 + (xfercount * xferwidth) as u32;
-            }
-            if dir == Direction::PeripheralToMemory {
-                DESCRIPTORS.list[channel].src_data_end_addr = srcbase as u32;
-            } else {
-                DESCRIPTORS.list[channel].src_data_end_addr = srcbase as u32 + (xfercount * xferwidth) as u32;
-            }
-            DESCRIPTORS.list[channel].nxt_desc_link_addr = 0;
-        }
-
         // Configure for transfer type, no hardware triggering (we'll trigger via software), high priority
         // SAFETY: unsafe due to .bits usage
         self.info.regs.channel(channel).cfg().write(|w| unsafe {
@@ -164,8 +146,19 @@ impl<'d> Channel<'d> {
         // SAFETY: unsafe due to .bits usage
         self.info.regs.channel(channel).xfercfg().write(|w| unsafe {
             w.cfgvalid().set_bit();
-            w.clrtrig().set_bit();
-            w.reload().clear_bit();
+
+            if options.is_continuous {
+                w.reload().enabled();
+                w.clrtrig().clear_bit();
+            } else {
+                w.reload().disabled();
+                w.clrtrig().set_bit();
+            }
+            if options.is_sw_trig {
+                w.swtrig().set_bit();
+            } else {
+                w.swtrig().clear_bit();
+            }
             w.setinta().set_bit();
             w.width().bits(options.width.into());
             if dir == Direction::PeripheralToMemory {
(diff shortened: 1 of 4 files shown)


src/dma/transfer.rs
```diff
@@ -16,13 +16,21 @@ pub struct TransferOptions {
 
     /// Transfer priority level
     pub priority: Priority,
+
+    /// Transfer is intended to be done continuously, such as for a circular buffer
+    pub is_continuous: bool,
+
+    /// Transfer is software triggerred
```
madeleyneVaca, line 23: nit: typo, triggered*

src/uart.rs
```diff
@@ -687,11 +700,82 @@ impl<'a> UartRx<'a, Async> {
...
+        T::Interrupt::unpend();
+        unsafe { T::Interrupt::enable() };
+
+        let rx_dma = dma::Dma::reserve_channel(rx_dma);
+        T::info().regs.fifocfg().modify(|_, w| w.dmarx().enabled());
+        // immediately configure and enable channel for circular buffered reception
+        rx_dma.configure_channel(
+            dma::transfer::Direction::PeripheralToMemory,
+            T::info().regs.fiford().as_ptr() as *const u8 as *const u32,
+            buffer as *mut [u8] as *mut u32,
```
madeleyneVaca, line 738: Can you add a comment here describing why you're casting twice like this? If I'm following, this looks like you've got potentially a reference to some bytes (slice?) for which you're getting a pointer (instead of the ref). And then you're converting that into a u32 pointer presumably by grouping the bytes?

src/uart.rs
```diff
@@ -758,6 +842,164 @@ impl<'a> UartRx<'a, Async> {
...
+
+                        // finally increment total bytes read
+                        bytes_read += read_len;
+                    }
+                }
+            } else {
+                // sleep until next polling epoch, or if we detect a UART transfer error event
+                let res = select(
+                    // use embassy_time to enable polling the bus for more data
+                    embassy_time::Timer::after_micros(buffer_config.polling_rate),
```
madeleyneVaca, line 951: FYI: our timer implementation is still based on the RTC so we can only sleep in increments of milliseconds. This will basically just be after_millis(1)

madeleyneVaca: if we're able to get the baudrate of the UART up to 4 mbps we'll likely need to dedicate some time to configuring the time_driver with a different, faster clock

src/uart.rs
```diff
@@ -758,6 +842,164 @@ impl<'a> UartRx<'a, Async> {
...
+                            Poll::Ready(Err(Error::Fail))
+                        } else {
+                            Poll::Pending
+                        }
+                    }),
+                )
+                .await;
+
+                match res {
+                    Either::First(()) | Either::Second(Ok(())) => (),
```
madeleyneVaca, line 996: do we need an on_drop() function in the event that we return an error from the second future?

src/uart.rs
```diff
@@ -218,10 +219,22 @@ impl<'a> UartTx<'a, Blocking> {
     }
 }
 
+struct BufferConfig {
+    #[cfg(feature = "time")]
```
JamesHuard, line 223: we should consider adding GH tasks/issues to track:
1. Using a different/more flexible/dedicated hardware timer for background polling of the UART DMA buffer
2. Refactoring the circular buffer to use one of the off-the-shelf embassy solutions if possible

src/dma/channel.rs
```diff
@@ -124,24 +124,6 @@ impl<'d> Channel<'d> {
...
-            } else {
-                DESCRIPTORS.list[channel].dst_data_end_addr = dstbase as u32 + (xfercount * xferwidth) as u32;
-            }
-            if dir == Direction::PeripheralToMemory {
-                DESCRIPTORS.list[channel].src_data_end_addr = srcbase as u32;
-            } else {
-                DESCRIPTORS.list[channel].src_data_end_addr = srcbase as u32 + (xfercount * xferwidth) as u32;
-            }
-            DESCRIPTORS.list[channel].nxt_desc_link_addr = 0;
-        }
```
felipebalbi, line 143: Should be part of its own commit. Why is this needed?

src/dma/channel.rs
```diff
@@ -164,8 +146,19 @@ impl<'d> Channel<'d> {
...
+                w.clrtrig().clear_bit();
+            } else {
+                w.reload().disabled();
+                w.clrtrig().set_bit();
+            }
+            if options.is_sw_trig {
+                w.swtrig().set_bit();
+            } else {
+                w.swtrig().clear_bit();
+            }
```
felipebalbi, line 161: how about using `variant()`?

```rust
w.reload().variant(options.is_continuous).
    clrtrig().variant(!options.is_continuous).
    swtrig().variant(options.is_sw_trig)
```

src/uart.rs
```diff
@@ -49,6 +49,7 @@ pub struct UartTx<'a, M: Mode> {
 /// Uart RX driver.
 pub struct UartRx<'a, M: Mode> {
     info: Info,
+    _buffer_config: Option<BufferConfig>,
```
felipebalbi, line 52: IMO, you should produce new structs that contain a `UartRx`. In fact, adding a new directory `uart` and move the current driver to `uart/mod.rs` would allow you to introduce all your new code within `uart/buffered.rs`. Then, you can have a new `struct BufferedUartRx`, `struct BufferedUartTx`, and `BufferedUart` all isolated from the regular, unbuffered uart. Additionally, you would need a `BufferConfig`. A regular `Config` would suffice, because you would already have a separate `buffered` module.

src/uart.rs
```diff
@@ -687,11 +700,82 @@ impl<'a> UartRx<'a, Async> {
...
+                self.read_buffered(buf).await
+            } else {
+                self.read_unbuffered(buf).await
+            }
+        }
+
+        #[cfg(not(feature = "time"))]
+        {
+            self.read_unbuffered(buf).await
+        }
```
felipebalbi, line 775: If you had your separate `buffered.rs` submodule, this wouldn't be necessary. You would have your isolated `read()` implementation.

felipebalbi (changes requested):

jeffglaum:
@huseinmsft is this still an active PR that you want to complete?

PR #321, opened 2025-02-19
Add support for FC14 SPI (Untested) and FC15 I2C

sukomath:
- Adding https://github.com/OpenDevicePartnership/embassy-imxrt/pull/235 to enable FC14 and FC15
- Changes in I2C module to support FC15 I2C master

@tullom FYI

### DIFF
--- a/src/flexcomm.rs
+++ b/src/flexcomm.rs
@@ -5,7 +5,9 @@ use paste::paste;
 
 use crate::clocks::{enable_and_reset, SysconPeripheral};
 use crate::pac;
-use crate::peripherals::{FLEXCOMM0, FLEXCOMM1, FLEXCOMM2, FLEXCOMM3, FLEXCOMM4, FLEXCOMM5, FLEXCOMM6, FLEXCOMM7};
+use crate::peripherals::{
+    FLEXCOMM0, FLEXCOMM1, FLEXCOMM14, FLEXCOMM15, FLEXCOMM2, FLEXCOMM3, FLEXCOMM4, FLEXCOMM5, FLEXCOMM6, FLEXCOMM7,
+};
 
 /// clock selection option
 #[derive(Copy, Clone, Debug)]
@@ -22,8 +24,17 @@ pub enum Clock {
     /// MASTER
     Master,
 
-    /// `FCn_FRG`
-    FcnFrg,
+    /// FCn_FRG with Main clock source
+    FcnFrgMain,
+
+    /// FCn_FRG with Pll clock source
+    FcnFrgPll,
+
+    /// FCn_FRG with Sfro clock source
+    FcnFrgSfro,
+
+    /// FCn_FRG with Ffro clock source
+    FcnFrgFfro,
 
     /// disabled
     None,
@@ -70,17 +81,20 @@ macro_rules! impl_flexcomm {
 			    Clock::Ffro => w.sel().ffro_clk(),
 			    Clock::AudioPll => w.sel().audio_pll_clk(),
 			    Clock::Master => w.sel().master_clk(),
-			    Clock::FcnFrg => w.sel().fcn_frg_clk(),
-			    Clock::None => w.sel().none(),
-			});
+			    Clock::FcnFrgMain => w.sel().fcn_frg_clk(),
+			    Clock::FcnFrgPll => w.sel().fcn_frg_clk(),
+			    Clock::FcnFrgSfro => w.sel().fcn_frg_clk(),
+			    Clock::FcnFrgFfro => w.sel().fcn_frg_clk(),
+			    Clock::None => w.sel().none(), // no clock? throw an error?
+            });
 			clkctl1.flexcomm($idx).frgclksel().write(|w| match clk {
-			    Clock::Sfro => w.sel().sfro_clk(),
-			    Clock::Ffro => w.sel().ffro_clk(),
-			    Clock::AudioPll => w.sel().frg_pll_clk(),
-			    Clock::Master => w.sel().main_clk(),
-			    Clock::FcnFrg => w.sel().frg_pll_clk(),
-			    Clock::None => w.sel().none(),
+			    Clock::FcnFrgMain => w.sel().main_clk(),
+			    Clock::FcnFrgPll => w.sel().frg_pll_clk(),
+			    Clock::FcnFrgSfro => w.sel().sfro_clk(),
+			    Clock::FcnFrgFfro => w.sel().ffro_clk(),
+				_ => w.sel().none(),    // not using frg ...
 			});
+            // todo: add support for frg div/mult
 			clkctl1
 			    .flexcomm($idx)
 			    .frgctl()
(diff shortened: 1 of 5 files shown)


sukomath:
@felipebalbi @jerrysxie @DM777-ms Adding FC15 support changes. For the review comment from https://github.com/OpenDevicePartnership/embassy-imxrt/pull/235 - are we planning to adding this in PAC? Let me know if there is anything related to this review comment that needs to be addressed in this PR. 

image

src/iopctl.rs
DM777-ms, line 1: Are the changes in iopctl.rs creating a "virtual" port 8 for FC15 I2C? What was the benefit for this solution?
Also, related PIOFC15_SCL,PIOFC15_SDA additions to the chip files.

src/i2c/mod.rs
```diff
@@ -122,9 +122,9 @@ macro_rules! impl_instance {
     };
 }
 
-impl_instance!(0, 1, 2, 3, 4, 5, 6, 7);
+impl_instance!(0, 1, 2, 3, 4, 5, 6, 7, 15);
 
-const I2C_COUNT: usize = 8;
+const I2C_COUNT: usize = 16;
```
felipebalbi, line 127: why 16? You added only one more. Maybe add a check somewhere to map FC15 to index 8?

src/iopctl.rs
```diff
@@ -575,3 +578,7 @@ impl_pin!(PIO7_28, 7, 28);
 impl_pin!(PIO7_29, 7, 29);
 impl_pin!(PIO7_30, 7, 30);
 impl_pin!(PIO7_31, 7, 31);
+
+// Use Port 8 for FC15 pins offset calculation
+impl_pin!(PIOFC15_SCL, 8, 0);
+impl_pin!(PIOFC15_SDA, 8, 1);
```
felipebalbi, line 584: Which part numbers have a port 8? I don't see a mention of Port 8 in UM11147.pdf

/me reads carefully

Oh, I see. There is no port 8, you're relying on that port not existing and just using the number. Instead, I think you need custom iopctl implemenation for FC15_I2C_SCL and FC15_I2C_SDA. Because these are dedicated pins, I don't think you should rely on `impl_pin!()` macro by passing a non-existent port number.

felipebalbi (changes requested):

src/iopctl.rs
sukomath: We need to configure FC15_I2C_SCL/SDA pins through IOPCTL by setting pull config, drive strength, slew rate etc for I2C usage. These are dedicated pins, so function need not be configured.  PIOFC15_SCL, PIOFC15_SDA were added to chip files so that these pins can be configured during Fuel Gauge I2C initialization. 

image

IOPCTL reg offset for FC15_I2C_SCL/SDA is at 0x400/0x404. I was trying to use the existing implementation in iopctl.rs to calculate reg offset for IOPCTL. For offset to be 0x400, port number has to be 8. 

    let offset = ((port as usize) << 7) + ((pin as usize) << 2);

Added a custom iopctl implementation for FC15 pins to avoid using port number 8

image

src/i2c/mod.rs
sukomath: Setting I2C_COUNT to 9, added check in index() to return index 8 for FC15

src/iopctl.rs
sukomath: Added iopctl implementation for FC15 pins
impl_FC15pin!(PIOFC15_SCL, 0);
impl_FC15pin!(PIOFC15_SDA, 1);

src/iopctl.rs
```diff
@@ -222,133 +231,254 @@ impl AnyPin {
...
+    /// and pin combination are not being used simultaneously.
+    ///
+    /// Failure to uphold these requirements will result in undefined behavior.
+    ///
+    /// See Table 297 in reference manual for a list of valid
+    /// pin and port number combinations.
+    #[must_use]
+    pub unsafe fn new(pin: u8) -> Self {
+        // Table 297:  FC15_I2C_SCL offset = 0x400, FC15_I2C_SCL offset = 0x404
+        let offset = 0x400 + ((pin as usize) << 2);
```
jerrysxie, line 260: Can we get this as an already defined constant from the PAC?

src/flexcomm.rs
```diff
@@ -98,6 +112,88 @@ macro_rules! impl_flexcomm {
 
 impl_flexcomm!(0, 1, 2, 3, 4, 5, 6, 7);
 
+// Add special case FLEXCOMM14
+impl sealed::Sealed for crate::peripherals::FLEXCOMM14 {}
```
jerrysxie, line 116: This PR seems to have changes for both FC14 and FC15? How well is the FC14 tested? Do we need to add similar pin constructs for FC14?

jerrysxie (changes requested):

src/iopctl.rs
sukomath: Following up on @felipebalbi 's comment from https://github.com/OpenDevicePartnership/mimxrt685s-pac/pull/13

I couldn't find the constant defined in pac. Can we use register accessor method fc15_i2c_scl() ?https://docs.rs/mimxrt685s-pac/latest/mimxrt685s_pac/iopctl/struct.RegisterBlock.html#method.fc15_i2c_scl

src/flexcomm.rs
sukomath: I haven't tested SPI on FC14. We may need changes added in spi driver for supporting FC14. If we want to add the changes for FC14 support only after testing SPI functionality, I can remove FC14 changes and add only FC15 for now. 

@DM777-ms Please add if you have done any testing for FC14

DM777-ms: No extensive testing yet. I ran FC14 I2C on a Minthos DK one day, attempting to read the UEFI spi flash ID register. The code did not crash, but all I read back was 0s. Needed more debug to figure out where a fix was needed. Could have been in configuration, uefi SPI flash enable pin, i2c, or fc14.

jerrysxie: @sukomath I am okay with either removing FC14 from this PR or update the commit message to include FC14. If you opt for the second approach, definitely add to the commit message and comment in the code to indicate that FC14 is not tested for our future selves.

src/iopctl.rs
jerrysxie: Yeah, the register accessor method or `as_ptr()` on the register for the pins should work fine. I cannot find a better way to get the constant either.

@felipebalbi is the register accessor method or `as _ptr()` the best way?

src/flexcomm.rs
sukomath: Updated commit message and added comment about FC14.

tullom (approved):

src/iopctl.rs
sukomath: Switched to using as_ptr()

src/flexcomm.rs
```diff
@@ -70,17 +81,20 @@ macro_rules! impl_flexcomm {
...
-			    Clock::Ffro => w.sel().ffro_clk(),
-			    Clock::AudioPll => w.sel().frg_pll_clk(),
-			    Clock::Master => w.sel().main_clk(),
-			    Clock::FcnFrg => w.sel().frg_pll_clk(),
-			    Clock::None => w.sel().none(),
+			    Clock::FcnFrgMain => w.sel().main_clk(),
+			    Clock::FcnFrgPll => w.sel().frg_pll_clk(),
+			    Clock::FcnFrgSfro => w.sel().sfro_clk(),
+			    Clock::FcnFrgFfro => w.sel().ffro_clk(),
+                _ => w.sel().none(),    // not using frg ...
```
jerrysxie, line 95: nit: Indentation seems to be off here.

src/iopctl.rs
```diff
@@ -222,133 +231,257 @@ impl AnyPin {
...
+    ///
+    /// Failure to uphold these requirements will result in undefined behavior.
+    ///
+    /// See Table 297 in reference manual for a list of valid
+    /// pin and port number combinations.
+    #[must_use]
+    pub unsafe fn new(pin: u8) -> Self {
+        // Table 297:  FC15_I2C_SCL offset = 0x400, FC15_I2C_SCL offset = 0x404
+        let iopctl = unsafe { crate::pac::Iopctl::steal() };
+        let mut reg = &*iopctl.fc15_i2c_scl().as_ptr().cast();
```
jerrysxie, line 261: nit: We can do this, then `reg` does not need to be mutable:
```
        let reg = if pin == 0 {
            &*iopctl.fc15_i2c_scl().as_ptr().cast()
        } else {
            &*iopctl.fc15_i2c_sda().as_ptr().cast()
        };
```

jerrysxie (approved):

src/flexcomm.rs
sukomath: Fixed indentation

src/iopctl.rs
sukomath: Added this change

src/iopctl.rs
sukomath: @felipebalbi Please help with reviewing the changes requested

DM777-ms (approved):

src/i2c/mod.rs
```diff
@@ -101,16 +101,24 @@ macro_rules! impl_instance {
...
+                let mut info_index = $n;
+                if $n == 15 {
+                    info_index = 8;
+                }
+
+                Info {
+                    regs: unsafe { &*crate::pac::[<I2c $n>]::ptr() },
+                    index: info_index,
+                }
+    		}
```
felipebalbi, line 113: Odd-looking indentation here. Are you sure you ran `rustfmt`?

src/i2c/mod.rs
```diff
@@ -101,16 +101,24 @@ macro_rules! impl_instance {
...
 
 		    #[inline]
 		    fn index() -> usize {
-			$n
+                if $n == 15 {
+                    return 8
+                }
+                $n
 		    }
 		}
```
felipebalbi, line 123: also odd indentation.

src/iopctl.rs
```diff
@@ -222,133 +231,258 @@ impl AnyPin {
...
+impl SealedPin for FC15Pin {}
+
+macro_rules! impl_iopctlpin {
+    ($pintype:ident) => {
+        impl IopctlPin for $pintype {
+            fn set_function(&self, function: Function) -> &Self {
+                match function {
+                    Function::F0 => {
+                        self.reg.modify(|_, w| w.fsel().function_0());
+                    }
```
felipebalbi, line 290: you can use `variant()`:

```rust
self.reg.modify(|_, w| w.fsel().variant(function))
```

No need for that `match` expression.

src/iopctl.rs
```diff
@@ -222,133 +231,258 @@ impl AnyPin {
...
+                    Pull::None => {
+                        self.reg.modify(|_, w| w.pupdena().disabled());
+                    }
+                    Pull::Up => {
+                        self.reg.modify(|_, w| w.pupdena().enabled().pupdsel().pull_up());
+                    }
+                    Pull::Down => {
+                        self.reg
+                            .modify(|_, w| w.pupdena().enabled().pupdsel().pull_down());
+                    }
```
felipebalbi, line 330: ditto WRT `variant()`

src/iopctl.rs
```diff
@@ -222,133 +231,258 @@ impl AnyPin {
...
+
+            fn set_slew_rate(&self, slew_rate: SlewRate) -> &Self {
+                match slew_rate {
+                    SlewRate::Standard => {
+                        self.reg.modify(|_, w| w.slewrate().normal());
+                    }
+                    SlewRate::Slow => {
+                        self.reg.modify(|_, w| w.slewrate().slow());
+                    }
+                }
```
felipebalbi, line 353: ditto

src/iopctl.rs
```diff
@@ -222,133 +231,258 @@ impl AnyPin {
...
+
+            fn set_drive_strength(&self, strength: DriveStrength) -> &Self {
+                match strength {
+                    DriveStrength::Normal => {
+                        self.reg.modify(|_, w| w.fulldrive().normal_drive());
+                    }
+                    DriveStrength::Full => {
+                        self.reg.modify(|_, w| w.fulldrive().full_drive());
+                    }
+                }
```
felipebalbi, line 365: ditto

src/iopctl.rs
```diff
@@ -222,133 +231,258 @@ impl AnyPin {
...
-                self.reg.modify(|_, w| w.pupdena().disabled());
+            fn set_drive_mode(&self, mode: DriveMode) -> &Self {
+                match mode {
+                    DriveMode::PushPull => {
+                        self.reg.modify(|_, w| w.odena().disabled());
+                    }
+                    DriveMode::OpenDrain => {
+                        self.reg.modify(|_, w| w.odena().enabled());
+                    }
+                }
```
felipebalbi, line 387: ditto

src/iopctl.rs
```diff
@@ -222,133 +231,258 @@ impl AnyPin {
...
+
+            fn set_input_inverter(&self, inverter: Inverter) -> &Self {
+                match inverter {
+                    Inverter::Disabled => {
+                        self.reg.modify(|_, w| w.iiena().disabled());
+                    }
+                    Inverter::Enabled => {
+                        self.reg.modify(|_, w| w.iiena().enabled());
+                    }
+                }
```
felipebalbi, line 399: ditto

felipebalbi (changes requested):

src/iopctl.rs
sukomath: I can't use variant(function) as argument for variant is of type 'Fsel' defined in pac while argument 'function' passed  set_function() is using type 'Function' defined in iopctl.rs.  Let me know if there is another way to simplify this. 

image

src/i2c/mod.rs
sukomath: I ran cargo +nightly fmt but it did not catch formatting errors. Does rustfmt not work in macros?(https://github.com/rust-lang/rustfmt/issues/8)

src/i2c/mod.rs
sukomath: Updated indentation manually. Running cargo +nightly fmt doesnt seem to catch this formatting error

src/iopctl.rs
sukomath: @felipebalbi I couldnt find a way to remove the match expression without having to change the arguments to `set_function()`. As `set_function()` is used in configuring many other peripherals(adc, gpio, uart,timer, i2c), we will need to add this change in multiple locations. Let me know if I missed something or if there is another way to remove `match `expression.

src/iopctl.rs
sukomath: Tracked in https://github.com/OpenDevicePartnership/embassy-imxrt/pull/321#discussion_r1994108299

src/iopctl.rs
sukomath: Tracked in https://github.com/OpenDevicePartnership/embassy-imxrt/pull/321#discussion_r1994108299

src/iopctl.rs
sukomath: Tracked in https://github.com/OpenDevicePartnership/embassy-imxrt/pull/321#discussion_r1994108299

src/iopctl.rs
sukomath: Tracked in https://github.com/OpenDevicePartnership/embassy-imxrt/pull/321#discussion_r1994108299

src/iopctl.rs
felipebalbi: seems like we're `Into<Fsel> for Function`. I'll add it later. Please, disregard this comment.

felipebalbi (approved):

PR #322, opened 2025-02-21
Changes to i2c master and slave to handle large data buffers

### LINKED ISSUE #255: I2C transaction needs to handle DMA transfer size large than 1024

akshat2112:
Since DMA only handles a maximum of 1024 bytes of data, i2c master and slave need to handle larger buffers by dividing them into chunks of smaller buffers.
The example also demonstrates how to handle both read and write operations for a large buffer.

Closes #255

### DIFF
--- a/examples/rt685s-evk/src/bin/i2c-loopback-largebuf.rs
+++ b/examples/rt685s-evk/src/bin/i2c-loopback-largebuf.rs
@@ -1,0 +1,124 @@
+#![no_std]
+#![no_main]
+
+extern crate embassy_imxrt_examples;
+
+use defmt::info;
+use embassy_executor::Spawner;
+use embassy_imxrt::i2c::master::{I2cMaster, Speed};
+use embassy_imxrt::i2c::slave::{Address, Command, I2cSlave, Response};
+use embassy_imxrt::i2c::{self, Async};
+use embassy_imxrt::{bind_interrupts, peripherals};
+use embedded_hal_async::i2c::I2c;
+
+const ADDR: u8 = 0x20;
+const MAX_I2C_CHUNK_SIZE: usize = 512;
+const MASTER_BUFLEN: usize = 2000;
+// slave buffer has to be 1 bigger than master buffer for each chunk
+//because master does not handle end of read properly
+const SLAVE_BUFLEN: usize = MASTER_BUFLEN + (MASTER_BUFLEN / MAX_I2C_CHUNK_SIZE) + 1;
+const SLAVE_ADDR: Option<Address> = Address::new(ADDR);
+
+bind_interrupts!(struct Irqs {
+    FLEXCOMM2 => i2c::InterruptHandler<peripherals::FLEXCOMM2>;
+    FLEXCOMM4 => i2c::InterruptHandler<peripherals::FLEXCOMM4>;
+});
+
+#[embassy_executor::task]
+async fn slave_service(mut slave: I2cSlave<'static, Async>) {
+    let mut r_buf = [0xAA; SLAVE_BUFLEN];
+    let mut t_buf = [0xAA; SLAVE_BUFLEN];
+
+    // Initialize write buffer with increment numbers
+    for (i, e) in t_buf.iter_mut().enumerate() {
+        *e = ((i / MAX_I2C_CHUNK_SIZE) as u8) + 1;
+    }
+    for (i, e) in r_buf.iter_mut().enumerate() {
+        *e = ((i as u8) % 255) + 1;
+    }
+
+    let mut r_offset = 0;
+    let mut t_offset = 0;
+
+    loop {
+        match slave.listen().await.unwrap() {
+            Command::Probe => {
+                info!("Probe, nothing to do");
+            }
+            Command::Read => {
+                info!("Read");
+                loop {
+                    let end = (t_offset + MAX_I2C_CHUNK_SIZE + 1).min(t_buf.len());
+                    let t_chunk = &t_buf[t_offset..end];
+                    match slave.respond_to_read(t_chunk).await.unwrap() {
+                        Response::Complete(n) => {
+                            t_offset += n;
+                            info!("Response complete read with {} bytes", n);
+                            break;
+                        }
+                        Response::Pending(n) => {
+                            t_offset += n;
+                            info!("Response to read got {} bytes, more bytes to fill", n);
+                        }
+                    }
+                }
+            }
+            Command::Write => {
+                info!("Write");
+                loop {
+                    let end = (r_offset + MAX_I2C_CHUNK_SIZE).min(r_buf.len());
+                    let r_chunk = &mut r_buf[r_offset..end];
+                    match slave.respond_to_write(r_chunk).await.unwrap() {
+                        Response::Complete(n) => {
+                            r_offset += n;
+                            if n == 0 {
+                                info!("Restart detected");
+                            } else {
+                                info!("Response complete write with {} bytes", n);
+                            }
+                            break;
+                        }
+                        Response::Pending(n) => {
+                            r_offset += n;
+                            info!("Response to write got {} bytes, more bytes pending", n);
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+
+#[embassy_executor::task]
+async fn master_service(mut master: I2cMaster<'static, Async>) {
+    const ADDR: u8 = 0x20;
+
+    let mut w_buf = [0xAA; MASTER_BUFLEN];
+    let mut r_buf = [0xAA; MASTER_BUFLEN];
+
+    // Initialize write buffer with increment numbers
+    for (i, e) in w_buf.iter_mut().enumerate() {
+        *e = ((i / MAX_I2C_CHUNK_SIZE) as u8) + 1;
+    }
+
+    let w_end = w_buf.len();
+    info!("i2cm write {} bytes", w_end);
+    master.write(ADDR, &w_buf[0..w_end]).await.unwrap();
+
+    let r_end = r_buf.len();
+    info!("i2cm read {} bytes", r_end);
+    master.read(ADDR, &mut r_buf[0..r_end]).await.unwrap();
+}
+
+#[embassy_executor::main]
+async fn main(spawner: Spawner) {
+    info!("i2c loopback bigbuffer example");
+    let p = embassy_imxrt::init(Default::default());
+
+    let slave = I2cSlave::new_async(p.FLEXCOMM2, p.PIO0_18, p.PIO0_17, Irqs, SLAVE_ADDR.unwrap(), p.DMA0_CH4).unwrap();
+
+    let master = I2cMaster::new_async(p.FLEXCOMM4, p.PIO0_29, p.PIO0_30, Irqs, Speed::Standard, p.DMA0_CH9).unwrap();
+
+    spawner.must_spawn(master_service(master));
+    spawner.must_spawn(slave_service(slave));
+}
--- a/src/i2c/master.rs
+++ b/src/i2c/master.rs
@@ -9,7 +9,7 @@ use embassy_hal_internal::into_ref;
 
 use super::{
     Async, Blocking, Error, Info, Instance, InterruptHandler, MasterDma, Mode, Result, SclPin, SdaPin, TransferError,
-    I2C_WAKERS,
+    I2C_WAKERS, MAX_I2C_CHUNK_SIZE,
 };
 use crate::interrupt::typelevel::Interrupt;
 use crate::{dma, interrupt, Peripheral};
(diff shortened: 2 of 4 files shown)


src/i2c/slave.rs
```diff
@@ -430,8 +448,26 @@ impl I2cSlave<'_, Async> {
 
     /// Respond to read command from master
     /// User must provide enough data to complete the transaction or else
-    ///    we will get stuck in this function
+    /// we will get stuck in this function
     pub async fn respond_to_read(&mut self, buf: &[u8]) -> Result<Response> {
+        let mut xfer_count = 0;
+        for chunk in buf.chunks(MAX_I2C_CHUNK_SIZE + 1) {
```
akshat2112, line 454: n+1 sized buffer is required for n byte read. Sending n+1 sized chunk causes n byte read and skips over one byte in the next chunk. Need to handle this properly

akshat2112: As per i2c spec the master should nack the last byte during read, signaling it has read the required number of bytes. On the RT685S-EVK, the master instead sends a nack on the byte _after_ the last required byte has been read. Is this a typical case that we need to handle here or should this be taken care by the caller of `respond_to_read`?

src/i2c/master.rs
```diff
@@ -661,18 +661,26 @@ impl embedded_hal_1::i2c::I2c for I2cMaster<'_, Blocking> {
...
-        self.read_no_stop(address, read).await?;
+        for chunk in read.chunks_mut(MAX_I2C_CHUNK_SIZE) {
+            self.read_no_stop(address, chunk).await?;
+        }
         self.stop().await
     }
 
     async fn write(&mut self, address: u8, write: &[u8]) -> Result<()> {
-        self.write_no_stop(address, write).await?;
+        for chunk in write.chunks(MAX_I2C_CHUNK_SIZE) {
```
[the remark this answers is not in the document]
jerrysxie: 0 byte read is not allowed in i2c spec but I believe 0 byte write is allowed.

jerrysxie (changes requested):

examples/rt685s-evk/src/bin/i2c-loopback-largebuf.rs
```diff
@@ -0,0 +1,124 @@
...
+        *e = ((i / MAX_I2C_CHUNK_SIZE) as u8) + 1;
+    }
+
+    let w_end = w_buf.len();
+    info!("i2cm write {} bytes", w_end);
+    master.write(ADDR, &w_buf[0..w_end]).await.unwrap();
+
+    let r_end = r_buf.len();
+    info!("i2cm read {} bytes", r_end);
+    master.read(ADDR, &mut r_buf[0..r_end]).await.unwrap();
```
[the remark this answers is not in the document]
akshat2112: Updated

examples/rt685s-evk/src/bin/i2c-loopback-largebuf.rs
```diff
@@ -0,0 +1,124 @@
...
+                        }
+                    }
+                }
+            }
+            Command::Write => {
+                info!("Write");
+                loop {
+                    let end = (r_offset + MAX_I2C_CHUNK_SIZE).min(r_buf.len());
+                    let r_chunk = &mut r_buf[r_offset..end];
+                    match slave.respond_to_write(r_chunk).await.unwrap() {
```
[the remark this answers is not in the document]
akshat2112: Updated

RobertZ2011 (dismissed):

jerrysxie (approved):

src/i2c/master.rs
```diff
@@ -761,6 +761,30 @@ impl<'a> I2cMaster<'a, Async> {
         )
         .await
     }
+
+    async fn write_chunks(&mut self, address: u16, write: &[u8]) -> Result<()> {
```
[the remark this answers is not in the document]
akshat2112: i2c write allows 0 byte write. Returning after the 0 byte write

src/i2c/slave.rs
```diff
@@ -502,6 +502,24 @@ impl I2cSlave<'_, Async> {
 
     /// Respond to write command from master
     pub async fn respond_to_write(&mut self, buf: &mut [u8]) -> Result<Response> {
```
akshat2112, line 504: @jerrysxie please double check if this is okay. With this change, the slave will not return `Response::Pending` to the caller. Instead, if we get `Response::Pending`, we try to read/write the next chunk of buffer. Based on the code the only possibility to get `Response::Pending` is when the master wants to write more data and has run out of buffer, so it makes sense. When master sends a STOP or when slave runs out of buffer chunks, we return `Response::Complete`

It works for the current examples we have, but want to make sure we're not missing out on any other cases

akshat2112: Further analysis: The case where slave runs out of buffer returns `Response::Complete`. This might be problematic since the caller would have no way to decipher if the write operation completed because the master sent all data or if the slave ran out of buffer

src/i2c/master.rs
```diff
@@ -761,6 +761,30 @@ impl<'a> I2cMaster<'a, Async> {
...
+        // write of 0 size is allowed according to i2c spec
+        if write.is_empty() {
+            self.write_no_stop(address, write).await?;
```
jerrysxie, line 768: Can we just return after this and not have to go into the for loop?

src/i2c/slave.rs
jerrysxie: @akshat2112 Next time if major changes are made in the PR, please reset my approval and re-request a review or else this shows as already reviewed and will be ignored by reviewers.

I approved this PR when it was only master changes, but then later on slave changes were added as well.

src/i2c/slave.rs
```diff
@@ -579,8 +597,26 @@ impl I2cSlave<'_, Async> {
 
     /// Respond to read command from master
     /// User must provide enough data to complete the transaction or else
-    ///    we will get stuck in this function
+    /// we will get stuck in this function
     pub async fn respond_to_read(&mut self, buf: &[u8]) -> Result<Response> {
+        let mut xfer_count = 0;
+        for chunk in buf.chunks(MAX_I2C_CHUNK_SIZE) {
+            let result = self.respond_to_read_inner(chunk).await?;
```
jerrysxie, line 604: This code will only work for the case where master sends a restart between chunks of data. It will not work if the master sends more than 1024 bytes of data without inserting a restart in-between.

src/i2c/slave.rs
jerrysxie: > Further analysis: The case where slave runs out of buffer returns `Response::Complete`. This might be problematic since the caller would have no way to decipher if the write operation completed because the master sent all data or if the slave ran out of buffer. Needs to be fixed

The case where the salve runs out of buffer returning `Response::Complete` is not a good behavior. That is not what the API is trying to provide to the client. `Pending` means there is more bytes that the master is sending, so the slave needs to read more. `Complete` means that the slave read all the bytes master sent and the master sent a STOP,

examples/rt685s-evk/.cargo/config.toml
```diff
@@ -2,16 +2,21 @@
...
+  "linker=flip-link",
+  "-C",
+  "link-arg=-Tlink.x",
+  "-C",
+  "link-arg=-Tdefmt.x",
   # This is needed if your flash or ram addresses are not aligned to 0x10000 in memory.x
   # See https://github.com/rust-embedded/cortex-m-quickstart/pull/95
-  "-C", "link-arg=--nmagic",
+  "-C",
+  "link-arg=--nmagic",
```
jerrysxie, line 14: Unnecessary format change.

jerrysxie (changes requested):
The slave portion of this will require more consideration.

src/i2c/mod.rs
```diff
@@ -127,6 +127,9 @@ impl_instance!(0, 1, 2, 3, 4, 5, 6, 7);
 const I2C_COUNT: usize = 8;
 static I2C_WAKERS: [AtomicWaker; I2C_COUNT] = [const { AtomicWaker::new() }; I2C_COUNT];
 
+/// Maximum I2C chunk size for DMA transfers
+pub const MAX_I2C_CHUNK_SIZE: usize = 1024;
```
felipebalbi, line 131: not part of this PR, but we should really define this within the DMA driver, as it's a DMA limitation :)

felipebalbi (approved):

felipebalbi (commented):
On comment only. I'll approve after Jerry's comments are resolved

felipebalbi (commented):
Made a mistake, see above

src/i2c/mod.rs
akshat2112: Do you suggest letting it be defined here as part of this PR and move it to the DMA driver as a separate change?

examples/rt685s-evk/.cargo/config.toml
akshat2112: Auto formatted due to extension Even Better TOML. Will revert in next update.

src/i2c/slave.rs
```diff
@@ -584,8 +602,26 @@ impl I2cSlave<'_, Async> {
...
+            match result {
+                Response::Complete(count) => {
+                    xfer_count += count;
+                }
+                Response::Pending(count) => {
+                    xfer_count += count;
+                    return Ok(Response::Pending(xfer_count));
```
akshat2112, line 616: @jerrysxie Which is the better approach to handling master reading more than the chunk size? I made it similar to the write operation, where the client can handle it by calling read again. But since we enforce the rule that client must provide enough data, we can also simply loop over the next chunk here instead of returning Pending to the client.

src/i2c/slave.rs
```diff
@@ -645,6 +686,9 @@ impl I2cSlave<'_, Async> {
             // then the next transaction starts the slave state goes into
             // pending + addressed.
             return Ok(Response::Complete(xfer_count));
+        } else if stat.slvstate().is_slave_transmit() {
```
jerrysxie, line 689: This is not a valid check due to NXP's I2C peripheral. The slvstate is only valid if pending bit is set. With DMA enabled, pending bit will not be set, so this is invalid. During my testing, I have found slvstate to not in transmitted even if master has more data to send.

src/i2c/slave.rs
jerrysxie: Due to race condition in hardware, currently `respond_to_read()` can never return pending, so unless we have a work around for the HW race condition in https://github.com/OpenDevicePartnership/embassy-imxrt/pull/322#discussion_r2021885563, we have to make the assumption that `respond_to_read()` has to complete the transaction. That is why the restriction is there in the first place.

examples/rt685s-evk/.cargo/config.toml
```diff
@@ -15,3 +15,4 @@ target = "thumbv8m.main-none-eabihf" # Cortex-M33
 
 [env]
 DEFMT_LOG = "trace"
+EMBASSY_EXECUTOR_TASK_ARENA_SIZE = "16384"
```
felipebalbi, line 18: Removed from upstream embassy, finally :)

src/i2c/mod.rs
```diff
@@ -135,6 +135,9 @@ impl_instance!(0, 1, 2, 3, 4, 5, 6, 7, 15);
 const I2C_COUNT: usize = 9;
 static I2C_WAKERS: [AtomicWaker; I2C_COUNT] = [const { AtomicWaker::new() }; I2C_COUNT];
 
+/// Maximum I2C chunk size for DMA transfers
+pub const MAX_I2C_CHUNK_SIZE: usize = 0x3FF;
```
felipebalbi, line 139: shouldn't this be defined in the DMA module as a `pub(crate)` constant?

Issue #170, opened 2024-11-04
Support Capture timers for Dual Edges of the input signal in the capture channel


Shaibal-Microsoft:
Right now, capture timer works for Rising and Falling edge of the input signal in the capture channel.
Dual edge requirements is not clear and hence the support was not added. Creating this bug for tracking the issue

PR #323, opened 2025-02-21
Update CaptureTimer to calculate fan speed

tim18763:
**Why?**
To calculate fan speed, we need to capture fan TACH signal period time, from this rising edge to next rising edge.

**What changes?**
1. Implement new function "capture_cycle_time_us" for CaptureTimer. This function trigger capture twice and return time us between these two captures.
2. Fix return value in "get_event_capture_time_us".

**How to test?**
Result from print log match with LA waveform

From print log
11:04:48.115 : (86918650) fan tach cycle 4355 us
11:04:48.147 : (86918650) calculated fan rpm = 6888

LA waveform of fan tach pin
image

#170

### DIFF
--- a/examples/rt685s-evk/src/bin/timer.rs
+++ b/examples/rt685s-evk/src/bin/timer.rs
@@ -47,15 +47,19 @@ async fn main(_spawner: Spawner) {
     // Creating a separate block to test Timer Drop logic
     {
         let sfro = ClockConfig::crystal().sfro;
-        let mut cap_async_tmr = CaptureTimer::new_async(p.CTIMER0_CAPTURE_CHANNEL0, sfro);
+        let mut cap_async_tmr = CaptureTimer::new_async(p.CTIMER0_CAPTURE_CHANNEL0, p.PIO1_7, sfro);
 
         // pass the input mux number, Input pin and Input pin edge user is interested in
         // Input mux details can be found in NXP user manual section 8.6.8 and Pin Function Table in section 7.5.3
-        let event_time_ms = cap_async_tmr
-            .capture_event_time_us(TriggerInput::TrigIn9, p.PIO1_7, CaptureChEdge::Falling)
-            .await;
+        let event_time_us = cap_async_tmr.capture_event_time_us(CaptureChEdge::Falling).await;
 
-        info!("Capture timer expired in = {} ms", event_time_ms);
+        info!("Capture timer expired in = {} us", event_time_us);
+
+        let sfro = ClockConfig::crystal().sfro;
+        let mut cap_async_tmr = CaptureTimer::new_async(p.CTIMER4_CAPTURE_CHANNEL0, p.PIO0_5, sfro);
+        let event_time_us = cap_async_tmr.capture_cycle_time_us(CaptureChEdge::Rising).await;
+
+        info!("Capture timer expired, time between two capture = {} us", event_time_us);
     }
 
     loop {
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -139,12 +139,13 @@ impl sealed::Sealed for Async {}
 impl Mode for Async {}
 
 /// A timer that captures events based on a specified edge and calls a user-defined callback.
-pub struct CaptureTimer<M: Mode> {
+pub struct CaptureTimer<M: Mode, P: CaptureEvent> {
     id: usize,
     event_clock_counts: u32,
     clk_freq: u32,
     _phantom: core::marker::PhantomData<M>,
     info: Info,
+    event_pin: P,
 }
 
 /// A timer that counts down to zero and calls a user-defined callback.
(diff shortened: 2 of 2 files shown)


examples/rt685s-evk/src/bin/timer.rs
```diff
@@ -51,11 +51,19 @@ async fn main(_spawner: Spawner) {
...
-        info!("Capture timer expired in = {} ms", event_time_ms);
+        info!("Capture timer expired in = {} us", event_time_us);
+
+        let sfro = ClockConfig::crystal().sfro;
+        let mut cap_async_tmr = CaptureTimer::new_async(p.CTIMER4_CAPTURE_CHANNEL0, sfro);
+        let event_time_us = cap_async_tmr
+            .capture_cycle_time_us(TriggerInput::TrigIn1, p.PIO0_5, CaptureChEdge::Rising)
+            .await;
+
+        info!("Capture timer expired, time between two capture = {} us", event_time_us);
```
tim18763, line 66: inherit conversation from #319 
Add usage example of capture_cycle_time_us

src/timer.rs
```diff
@@ -435,10 +435,8 @@ impl<M: Mode> CaptureTimer<M> {
     /// Captured clock = (Capture value - previous counter value)
     fn get_event_capture_time_us(&self) -> u32 {
         let time_float = (self.event_clock_counts as f32 / self.clk_freq as f32) * 1000000.0;
-        let time_int = time_float as u32 as f32;
-        let interger_part = time_int as u32;
-        let decimal_part = ((time_float - time_int) * 1000000.0) as u32;
-        interger_part + decimal_part
+        let integer_part = time_float as u32;
+        integer_part
```
tim18763, line 439: inherit conversation from #319 
image

take this as example:
if time_float is number like "1234.5678"us
integer_part will be "1234", but decimal_part will be "567800" because it's multiply 1000000 in removed line 440.
plus them and return result will be "569034"us which not correct
 
If we want to keep decimal part, maybe we can return type f32?

src/timer.rs
```diff
@@ -532,6 +530,49 @@ impl CaptureTimer<Async> {
         })
         .await
     }
+
+    /// Trigger capture twice, return time us between these two capture
+    /// TODO: https://github.com/OpenDevicePartnership/embassy-imxrt/issues/229
```
tim18763, line 535: inherit conversation from #319 
image

Add comment TODO and link the bug for future enhancement.

src/timer.rs
```diff
@@ -532,6 +530,49 @@ impl CaptureTimer<Async> {
...
+        let mut first_captured = false;
+
+        // Implementation of waiting for the interrupt
+        poll_fn(|cx| {
+            WAKERS[self.id].register(cx.waker());
+
+            if self.info.input_event_captured() {
+                // First time capture, store data into timer hist and reenable interrupt
+                if first_captured == false {
+                    timer_hist = reg.cr(self.info.channel).read().bits();
```
tim18763, line 555: inherit conversation from #319 
image

Fixed, use flag first_captured to identify first capture or second capture

tim18763:
It's a clean commit history PR from #319 .

src/timer.rs
```diff
@@ -574,6 +616,40 @@ impl CaptureTimer<Blocking> {
             }
         }
     }
+    /// Trigger capture twice, return time us between these two capture
+    pub fn capture_cycle_time_us(
```
madeleyneVaca, line 620: Instead of this way of passing the pin peripheral to this function, can we instead add the tach pin to the CaptureTimer struct? This approach makes us have to clone the pin each time we want to get the fan speed. That's not the safe/idiomatic way to go.

madeleyneVaca (changes requested):

src/timer.rs
tim18763: PR updated. Add event_pin for CaptureTimer struct. And need to assign pin when creating CaptureTimer by new_async or new_blocking.

madeleyneVaca (approved):

src/timer.rs
```diff
@@ -483,9 +482,9 @@ impl<M: Mode> CaptureTimer<M> {
     }
 }
 
-impl CaptureTimer<Async> {
+impl<P: CaptureEvent> CaptureTimer<Async, P> {
     /// Creates a new `CaptureTimer` in asynchronous mode.
-    pub fn new_async<T: Instance>(_inst: T, clk: impl ConfigurableClock) -> Self {
+    pub fn new_async<T: Instance>(_inst: T, pin: P, clk: impl ConfigurableClock) -> Self {
```
sukomath, line 487: It would be good to add a trait bound for 'impl CaptureEvent' on pin. This will ensure that only CaptureEvent pins are used.

madeleyneVaca: P is required to implement CaptureEvent by the typestate so he doesn't need `impl CaptureEvent` in the constructor since it'd be redundant.

src/timer.rs
jerrysxie: Good catch, I got confused by this math even in the original review.

jerrysxie (approved):
Is there plan to work on adding DMA support? So we won't miss an edge: https://github.com/OpenDevicePartnership/embassy-imxrt/pull/319#discussion_r1964608481?

PR #325, opened 2025-02-24
espi: use variant() where applicable

felipebalbi:
Instead of explicit if-else expressions, we can simplify the code by using `variant()` whenever possible.


RobertZ2011 (approved):

akshat2112 (approved):

felipebalbi:
ping

jerrysxie (approved):

PR #332, opened 2025-02-27
Optimize i2c read by handling the last byte manually

### LINKED ISSUE #304: For buffer of size N, i2c master as a receiver for an I2c read should send a NACK on N byte instead of N + 1 byte
Normally a slave will send dummy byte until the master sends a NACK. However, it seems like some i2c slaves are not tolerant of this.

akshat2112:
For buffer of size N, i2c master sends a NACK on (N+1)th byte instead of Nth byte. This change separates the read operation into two steps. The first N-1 bytes are read using DMA and the last byte is read manually, which generates the expected NACK at Nth byte.
For a single byte read operation, the DMA is bypassed and the byte is read manually.

Closes #304

### DIFF
--- a/src/i2c/master.rs
+++ b/src/i2c/master.rs
@@ -372,52 +372,57 @@ impl<'a> I2cMaster<'a, Async> {
 
         self.start(address, true).await?;
 
-        let transfer = dma::transfer::Transfer::new_read(
-            self.dma_ch.as_mut().unwrap(),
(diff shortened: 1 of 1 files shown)


jerrysxie (approved):

jerrysxie:
Great work!

tullom (approved):
good handling of edge cases 😄

PR #333, opened 2025-02-28
Add NoDma I2C master async bus implementation

### LINKED ISSUE #327: Implement NoDma for I2C Master
Flexcomm 15 does not have an associated DMA channel. With the current implementation of Async I2C Master (which requires a DMA channel for async functionality), there is no way to run the I2C Master on Flexcomm 15 asynchronously.

This issue should be closed when an async I2C master without DMA is implemented, allowing Flexcomm 15 (as well as the other DMA) to run asynchronously without DMA.

tullom:
Closes #327 

Some kicking off questions:
- The reason I need to make a `new_async_nodma()` fn instead of hijacking `new_async()` is because `new_async()` requires a dma_ch object that impls Peripheral and impls an interrupt with `MasterDma`. I tried creating structs to impl all that but I hit a wall.
- I added documentation saying NOT to use this and to prefer `new_async()`, except in the cases of the Flexcomm not having DMA support. Is this enough or do you think this should be gated behind some sort of feature flag to reduce the possibility of misuse?

EDIT: 
I was able to create a NoDma structure. I had to minimally modify the DMA api, specifically `reserve_channel()` now returns an option that will return None if NoDma is used.

### DIFF
--- a/src/hashcrypt/mod.rs
+++ b/src/hashcrypt/mod.rs
@@ -104,7 +104,7 @@ impl<'d> Hashcrypt<'d, Async> {
         peripheral: impl Peripheral<P = HASHCRYPT> + 'd,
         dma_ch: impl Peripheral<P = impl HashcryptDma> + 'd,
     ) -> Self {
-        Self::new_inner(peripheral, Some(dma::Dma::reserve_channel(dma_ch)))
+        Self::new_inner(peripheral, dma::Dma::reserve_channel(dma_ch))
     }
 
     /// Start a new SHA256 hash
--- a/src/i2c/master.rs
+++ b/src/i2c/master.rs
@@ -331,7 +331,7 @@ impl<'a> I2cMaster<'a, Async> {
         T::into_i2c();
 
         let ch = dma::Dma::reserve_channel(dma_ch);
-        let this = Self::new_inner::<T>(fc, scl, sda, speed, Some(ch))?;
+        let this = Self::new_inner::<T>(fc, scl, sda, speed, ch)?;
 
         T::Interrupt::unpend();
         unsafe { T::Interrupt::enable() };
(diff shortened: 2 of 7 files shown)


src/i2c/master.rs
```diff
@@ -88,6 +90,12 @@ impl<'a, M: Mode> I2cMaster<'a, M> {
             // SAFETY: only unsafe due to .bits usage
             unsafe { w.mstsclhigh().bits(0).mstscllow().bits(1) });
 
+        if timeout.is_some() && timeout.unwrap() {
```
jerrysxie, line 93: We should remove the timeout option as we are not utilizing them correctly. If we want to apply this, then in our poll_fn() and in our interrupt handler, we need to check for the timeout condition and clean up transaction accordingly.

src/i2c/master.rs
```diff
@@ -262,7 +283,35 @@ impl<'a> I2cMaster<'a, Async> {
...
+        Ok(this)
+    }
+
+    /// IMPORTANT: DO NOT USE unless you are aware of the performance implications of not using DMA.
+    /// This NoDma I2C bus should only be used when a Flexcomm doesn't support DMA, such as Flexcomm 15.
+    /// This bus performs byte by byte transfers, awaiting between each byte,
+    /// which can lead to bus timeouts if a byte is not handled fast enough.
+    ///
+    /// use flexcomm fc with Pins scl, sda as an I2C Master bus, configuring to speed and pull
+    pub fn new_async_nodma<T: Instance>(
```
jerrysxie, line 300: Instead of implementing this for all the instances, can we also only implement this for the I2C instance corresponding to FC15?

jerrysxie (changes requested):

src/i2c/master.rs
felipebalbi: I agree. In fact, it took some work to remove all the previous `timeout` behavior we had. Rather not have it again.

src/i2c/master.rs
felipebalbi: Why? User can decide to not use DMA on any instance, right? Should we make the choice for them? I'm not convinced, IMO.

In any case, we shouldn't add a new method here. Just patch `new_async()` to take an `Option<impl Peripheral...>` or something like that. Could you give that a go?

felipebalbi (changes requested):

src/i2c/master.rs
tullom: So my issue with patching `new_async()` is that that would be a breaking change to our I2C driver. If we are ok with that, I can do it (and would prefer it if we aren't going the FC15 only route)

src/i2c/master.rs
tullom: Understood, will remove.

tullom: Removed

src/i2c/master.rs
tullom: FYI @jerrysxie @felipebalbi
I was able to create a NoDma structure. I had to minimally modify the DMA api, specifically `reserve_channel()` now returns an option that will return `None` if NoDma is used. Needed to slightly modify the UART driver and hashcrypt driver as well to return errors if NoDma is used since it doesn't make sense if NoDma is used.

Tested all touched drivers via examples on hardware and it works flawlessly.

RobertZ2011 (approved):

src/i2c/mod.rs
```diff
@@ -332,3 +332,20 @@ impl_dma!(FLEXCOMM6, Master, DMA0_CH13);
...
+}
+
+impl_nodma!(FLEXCOMM0, Master);
+impl_nodma!(FLEXCOMM1, Master);
+impl_nodma!(FLEXCOMM2, Master);
+impl_nodma!(FLEXCOMM3, Master);
+impl_nodma!(FLEXCOMM4, Master);
+impl_nodma!(FLEXCOMM5, Master);
+impl_nodma!(FLEXCOMM6, Master);
+impl_nodma!(FLEXCOMM7, Master);
```
sukomath, line 351: Shouldn't we add FLEXCOMM14 and 15 to this list?

tullom: In order to impl the `MasterDma` trait, flexcomm 15 needs impl the i2c::Instance trait, which i see you are adding support for in this PR: https://github.com/OpenDevicePartnership/embassy-imxrt/pull/321

Flexcomm 14 should not be added to the list since it is a SPI only flexcomm and this is the i2c module.

src/hashcrypt/mod.rs
```diff
@@ -103,8 +111,13 @@ impl<'d> Hashcrypt<'d, Async> {
...
         dma_ch: impl Peripheral<P = impl HashcryptDma> + 'd,
-    ) -> Self {
-        Self::new_inner(peripheral, Some(dma::Dma::reserve_channel(dma_ch)))
+    ) -> Result<Self, Error> {
+        let ch = dma::Dma::reserve_channel(dma_ch);
+        if ch.is_some() {
+            Ok(Self::new_inner(peripheral, ch))
+        } else {
+            Err(Error::UnsupportedConfiguration)
+        }
```
felipebalbi, line 120: do you need this here, though? Only the actual channel implements `HashcryptDma`, so there's no way to pass `NoDma` here, meaning that `reserve_channel()` should always work, right?

tullom: Oh you are totally right. Went back and updated this for both hashcrypt and uart, as it is impossible to use NoDma with those drivers.

felipebalbi (approved):

src/i2c/mod.rs
tullom: #321 merged, Flexcomm 15 NoDma support added.

jerrysxie (approved):

PR #339, opened 2025-03-11
espi: set SRST bit on port complete

felipebalbi:
SRST appears to be required on port complete as, without it, transfers get wedged.


jerrysxie (approved):

tullom (approved):
Approved, just make sure that the deny workflow issue is resolved either thru an exception (another PR) or something else

PR #341, opened 2025-03-14
Espi VWires

felipebalbi:
Expose VWire API for the application.

### DIFF
--- a/examples/rt633/src/bin/espi.rs
+++ b/examples/rt633/src/bin/espi.rs
@@ -9,7 +9,7 @@ use defmt::{error, info};
 use embassy_executor::Spawner;
 use embassy_imxrt::bind_interrupts;
 use embassy_imxrt::espi::{
-    Base, Capabilities, Config, Direction, Espi, Event, InterruptHandler, Len, Maxspd, PortConfig,
+    Base, BootStatus, Capabilities, Config, Direction, Espi, Event, InterruptHandler, Len, Maxspd, PortConfig,
 };
 use embassy_imxrt::peripherals::ESPI;
 use {defmt_rtt as _, panic_probe as _};
@@ -77,6 +77,10 @@ async fn main(_spawner: Spawner) {
 
     data.fill(0);
 
+    // Boot success
+    espi.boot_status(BootStatus::Success);
+    espi.boot_done();
+
     loop {
         let event = espi.wait_for_event().await;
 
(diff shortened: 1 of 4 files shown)


examples/rt633/src/bin/espi.rs
```diff
@@ -77,6 +77,10 @@ async fn main(_spawner: Spawner) {
 
     data.fill(0);
 
+    // Boot success
+    espi.boot_status(BootStatus::Success);
+    espi.boot_done();
```
jerrysxie, line 82: Does doing this clear the PLRTSTN bit in WIRERO register?

src/espi.rs
```diff
@@ -617,6 +751,113 @@ impl<'d> Espi<'d> {
...
+    pub fn host_reset_ack(&mut self) {
+        self.info.regs.wirewo().write(|w| w.host_rst_ack().set_bit());
+        self.block_for_vwire_done();
+    }
+
+    /// Acknowledge Suspend Warn.
+    ///
+    /// Active Low.
+    pub fn suspend_ack(&mut self) {
+        self.info.regs.wirewo().write(|w| w.susackn().clear_bit());
```
jerrysxie, line 821: Is this active low or active high? The manual seems to claim set to ack:
image

jerrysxie (changes requested):

src/espi.rs
felipebalbi: N at the means usually means "negated", or active low.

src/espi.rs
```diff
@@ -617,6 +751,113 @@ impl<'d> Espi<'d> {
...
+
+    /// To be called when Host goes into G3.
+    ///
+    /// Active High.
+    pub fn dsw_pwrok_reset(&mut self) {
+        self.info.regs.wirewo().write(|w| w.dsw_pwrok_rst().set_bit());
+        self.block_for_vwire_done();
+    }
+
+    fn block_for_vwire_done(&self) {
```
tullom, line 856: Nit: maybe some documentation should be added to the public functions that call this function to let users know this fn busy loops and could take a while.

src/timer.rs
```diff
@@ -909,7 +909,7 @@ impl embedded_hal_02::Pwm for CTimerPwm<'_> {
                 reg.emr().modify(|_, w| w.emc1().set_());
 
                 // Write 1 to IR bit to clear interrupt
-                reg.ir().modify(|_, w| w.mr1int().set_bit());
```
tullom, line 912: Looks like you're clearing the bit instead of setting it, do you need to update the comment?

tullom (changes requested):

src/timer.rs
felipebalbi: This is a W1C bit. Write one to clear. @jerrysxie patched the PACs to provide the proper accessors for these timer bits, but HAL was left behind.

examples/rt633/src/bin/espi.rs
felipebalbi: Documentation doesn't say anything about that.

src/espi.rs
felipebalbi: done

src/espi.rs
jerrysxie: Yeah, on schematic, here, I am not so sure, but oh well, we can always change it later. Not enough documentation to tell.

jerrysxie (approved):

tullom (approved):

PR #344, opened 2025-03-18
eSPI mailbox split

felipebalbi:
Add support for Mailbox Split port configuration.

### DIFF
--- a/src/espi.rs
+++ b/src/espi.rs
@@ -553,7 +553,7 @@ impl<'d> Espi<'d> {
                 offset,
                 length,
             } => {
-                self.mailbox(port, direction, addr, offset, length);
+                self.mailbox(port, config.into(), direction, addr, offset, length);
             }
 
             PortConfig::MailboxSingle {
@@ -562,7 +562,7 @@ impl<'d> Espi<'d> {
                 offset,
                 length,
             } => {
-                self.mailbox_single(port, direction, addr, offset, length);
+                self.mailbox(port, config.into(), direction, addr, offset, length);
             }
 
             _ => {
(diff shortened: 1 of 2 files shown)


philgweber (approved):

src/espi.rs
```diff
@@ -562,7 +576,16 @@ impl<'d> Espi<'d> {
                 offset,
                 length,
             } => {
-                self.mailbox_single(port, direction, addr, offset, length);
+                self.mailbox(port, config.into(), direction, addr, offset, length);
+            }
+
+            PortConfig::MailboxSplit {
```
jerrysxie, line 582: Can we add an usage example for this? I imageine that the client will have to do this:
```
    let mut espi = Espi::new(
        p.ESPI,
        p.PIO7_29,
        p.PIO7_26,
        p.PIO7_27,
        p.PIO7_28,
        p.PIO7_30,
        p.PIO7_31,
        p.PIO7_25,
        p.PIO7_24,
        Irqs,
        Config {
            caps: Capabilities {
                max_speed: Maxspd::SmallThan20m,
                alert_as_a_pin: true,
                ..Default::default()
            },
            ram_base: 0x2000_0000,
            base0_addr: 0x2002_0000,
            base1_addr: 0x2003_0000,
            status_addr: Some(0x480),
            status_base: Base::OffsetFrom0,
            ports_config: [
                PortConfig::MailboxSplit { 
                    direction: Direction::IgnoreWriteReadOnly,
                    addr: 0,
                    offset: 0,
                    length: Len::Len256,
                },
                PortConfig::MailboxSplit { 
                    direction: Direction::IgnoreReadWriteOnly,
                    addr: 0,
                    offset: 256,
                    length: Len::Len256,
                },
                Default::default(),
                Default::default(),
                Default::default(),
            ],
            ..Default::default()
        },
    );
```
```
        match event {
            Ok(Event::Port0(_port_event)) => {
                info!("Port 0: Contents: {:08x}", &data[..64]);
                espi.complete_port(0).await;
            }
            Ok(Event::Port1(_port_event)) => {
                info!("Port 1: Contents: {:08x}", &data[..64]);
                espi.complete_port(1).await;
            }
```

jerrysxie (approved):

src/espi.rs
felipebalbi: Right, I can add, but it's the same as the existing example. All mailbox ports are configured and used (almost) exactly the same.

jerrysxie: We can either update the existing example or add some comment to guide the usage.

felipebalbi: Updated existing. Note that we don't need two ports, only one port is enough :-) One port handles both directions.

jerrysxie: Thanks, that is actually good information. If we are using one port, then how does it work? The event would come from the same port, but ram length is x2. But is write only/read only access enforced? Do we need to use 2 ports to enforce that only one area is read-only and the other area is write-only?

felipebalbi:
ping

src/espi.rs
jerrysxie: @felipebalbi How do we deal with this case?

jerrysxie (approved):

tullom (approved):

PR #355, opened 2025-03-24
Ensure GPIO pins are reset when used by a I2C {master, slave}, UART and ADC

Wassasin:
As the first step towards implementing `Drop` for all drivers, have the drivers own an instance of a `Pin` driver which takes care of resetting each pin when the parent peripheral driver is dropped.

A subsequent PR will tackle Flexcomm disable/reset for each peripheral.

I took care to change as little as possible. Ideally low level `Sealed` implementations are separated from high level embassy driver implementations. Currently this distinction is a bit blurry, and consequently a driver can be a bit messy. Good examples of these are the `SdaPin` and `CtsPin` traits and friends. The separation between `gpio` and `iopctl` could also be better. `IopctlPin` traits should be private.

Have not yet tested this PR on the EVK, as we are waiting for those to be delivered. I will update this PR once it has been tested.

### DIFF
--- a/examples/rt685s-evk/src/bin/clocks-blinky.rs
+++ b/examples/rt685s-evk/src/bin/clocks-blinky.rs
@@ -5,7 +5,7 @@ extern crate embassy_imxrt_examples;
 
 use defmt::{error, info};
 use embassy_executor::Spawner;
-use embassy_imxrt::iopctl::IopctlPin;
+use embassy_imxrt::iopctl::{IopctlFunctionPin, IopctlPin};
 use embassy_imxrt::{clocks, gpio};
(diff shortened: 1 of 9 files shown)


jamesmunns (approved):

jeffglaum:
@JamesHuard , @felipebalbi can you guys please review and provide feedback? Thanks.

jeffglaum:
@RobertZ2011 can you guys please review and provide feedback? Thanks.

felipebalbi (approved):
No worries with this change. Just added a Guard type to make sure `Drop` puts the pin back into its reset state. Conflicts need to be fixed, though.

Issue #80, opened 2024-09-18
Additional Flexcomm features


DM777-ms:
Additional flexcomm work needed to align with legacy application requirements, and enable peripheral features:
-Disable fc channel (for low power sleep/shutdown)
-Set FRG fractional clock DIV/MULT/SOURCE (TBD if needed - unused on previous gen non-Rust mcu)
-Lock FC channel (TBD if needed - unused on previous gen non-Rust mcu)

jerrysxie:
Not relevant at this point, closed.

PR #356, opened 2025-03-25
Adapted Flexcomm to a refcounted generic instance

Wassasin:
Uart and I2C, peripherals based on Flexcomm, clean up their Flexcomm instance when dropped.

Partially implements point 1. in #80. Briefly tested on rt685s-evk UART TX.

### DIFF
--- a/src/flexcomm.rs
+++ b/src/flexcomm.rs
@@ -1,8 +1,10 @@
 //! implements flexcomm interface wrapper for easier usage across modules
 
+use core::sync::atomic::{AtomicU8, Ordering};
+
 use paste::paste;
 
-use crate::clocks::{enable_and_reset, SysconPeripheral};
+use crate::clocks::{disable, enable_and_reset, SysconPeripheral};
 use crate::peripherals::{
     FLEXCOMM0, FLEXCOMM1, FLEXCOMM14, FLEXCOMM15, FLEXCOMM2, FLEXCOMM3, FLEXCOMM4, FLEXCOMM5, FLEXCOMM6, FLEXCOMM7,
 };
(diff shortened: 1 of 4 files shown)


src/flexcomm.rs
```diff
@@ -103,7 +163,23 @@ macro_rules! impl_flexcomm {
...
+            clkctl1.flexcomm($idx).fcfclksel().write(|w| w.sel().none());
+            clkctl1.flexcomm($idx).frgclksel().write(|w| w.sel().none());
+            disable::<[<FLEXCOMM $idx>]>();
+            }
+
+            #[allow(private_interfaces)]
+            fn state() -> &'static State {
+            static STATE: State = State::new();
+            &STATE
+            }
```
felipebalbi, line 182: please run `cargo +nightly fmt` on your code.

felipebalbi (changes requested):

src/flexcomm.rs
Wassasin: For me `cargo +nightly fmt` does not touch this code. Because this segment is in a `paste!` block, `fmt` seems to struggle with it. Honestly the original code also was formatted incorrectly, and I mimicked it thinking there was maybe a good reason for it.

If for you `fmt` works fine, please let me know. Meanwhile I'll fix the formatting by hand.

diondokter:
@Wassasin you'll need to squash your commits because all commits must pass CI

src/flexcomm.rs
```diff
@@ -45,66 +47,140 @@ mod sealed {
     pub trait Sealed {}
 }
 
+struct State {
+    refcount: AtomicU8,
+}
+
+impl State {
```
RobertZ2011, line 54: @tullom I think you were looking at GPIO ref counting at some point. Did that code get in? I didn't immediately see it, but I wanted to check if we already had a ref counting implementation. If we don't already have an implementation, then can some of the ref counting code be made generic and placed in its own module?

src/flexcomm.rs
```diff
@@ -45,66 +47,140 @@ mod sealed {
...
+        }
+    }
+}
+
+/// A generic reference to a usage of a Flexcomm peripheral.
+///
+/// Can be cloned to share the usage of the peripheral across multiple sites.
+/// Dropping the last reference will disable the Flexcomm peripheral.
+#[must_use]
+pub(crate) struct FlexcommRef {
```
RobertZ2011, line 67: Just to check, does this change any externally visible APIs?

RobertZ2011 (changes requested):

src/flexcomm.rs
Wassasin: It does not, this is completely internal. The only difference is that now for some Peripherals a non-trivial `Drop` implementation exists.

jerrysxie (approved):

jamesmunns (approved):

felipebalbi (approved):

src/flexcomm.rs
Wassasin: That PR was closed without merging @ https://github.com/OpenDevicePartnership/embassy-imxrt/pull/139 .

tullom (approved):

jerrysxie (approved):

RobertZ2011 (approved):

PR #357, opened 2025-03-25
Fix cannot set UART baudrate to those value other than 115200 issue

xushengfei:
In current Uart code, it only accepts 115200. If we set it to 3000000 or 4000000, the whole system will crash.
We need 3 changes to fix it:

Open clock setting to have user to set different type, for example: FFRO. Now it is hardcoded as SFRO.
Open source clock setting in Hz, we need to set it as 48M, but it is hardcoded as 16M Hz.
There is a bug in current code to calculate brgval.
let brgval = (source_clock_hz / ((osrval + 1) * baudrate)) - 1;
if source_clock_hz is 48M and baudrate is 4M, osrval is 15, brgval is be -1, since the type is u32, it cannot be accept by rust, it will return error.
Need to use compare the number instead of compare it with FFFF.
I tested it in my device by setting 4M baudrarte, it can work as expected.

### DIFF
--- a/src/uart.rs
+++ b/src/uart.rs
@@ -556,7 +556,7 @@ impl<'a> UartTx<'a, Async> {
 
         let tx_dma = dma::Dma::reserve_channel(tx_dma);
 
-        Ok(Self::new_inner::<T>(Some(tx_dma)))
+        Ok(Self::new_inner::<T>(tx_dma))
     }
 
     /// Transmit the provided buffer asynchronously.
@@ -687,7 +687,7 @@ impl<'a> UartRx<'a, Async> {
 
         let rx_dma = dma::Dma::reserve_channel(rx_dma);
 
-        Ok(Self::new_inner::<T>(Some(rx_dma)))
+        Ok(Self::new_inner::<T>(rx_dma))
     }
 
     /// Read from UART RX asynchronously.
@@ -788,8 +788,8 @@ impl<'a> Uart<'a, Async> {
 
         Ok(Self {
             info: T::info(),
-            tx: UartTx::new_inner::<T>(Some(tx_dma)),
-            rx: UartRx::new_inner::<T>(Some(rx_dma)),
+            tx: UartTx::new_inner::<T>(tx_dma),
+            rx: UartRx::new_inner::<T>(rx_dma),
         })
     }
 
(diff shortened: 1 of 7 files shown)


huseinmsft (approved):

src/uart.rs
```diff
@@ -371,12 +366,11 @@ impl<'a, M: Mode> Uart<'a, M> {
...
                 (u32::MAX, u32::MAX, u32::MAX),
                 |(best_diff, best_osr, best_brg), osrval| {
-                    let brgval = (source_clock_hz / ((osrval + 1) * baudrate)) - 1;
-                    let diff;
-
-                    if brgval > 65535 {
+                    if source_clock_hz < ((osrval + 1) * baudrate) {
                         (best_diff, best_osr, best_brg)
                     } else {
+                        let brgval = (source_clock_hz / ((osrval + 1) * baudrate)) - 1;
```
jerrysxie, line 372: The brgval overflow check is still necessary as the hardware register is only 16-bits:
image

src/uart.rs
```diff
@@ -332,22 +336,13 @@ impl<'a, M: Mode> Uart<'a, M> {
...
         }
 
-        Self::set_baudrate_inner::<T>(config.baudrate)?;
+        Self::set_baudrate_inner::<T>(config.baudrate, config.source_clock_hz)?;
         Self::set_uart_config::<T>(config);
 
         Ok(())
     }
 
-    fn get_fc_freq() -> u32 {
```
jerrysxie, line 341: I don't think we should allow the user to pass in the clock rate for sfro and ffro as they are static clocks. We should add support to get the rate from the clock module ideally.

For now, but keep this function and make a match statement that only supports sfro and ffro:

```rust
match {
sfro => 16_000_000,
ffro => 48_000_000,
_ => //return an error,
}
```

jerrysxie (changes requested):

JamesHuard (approved):

jerrysxie:
> Hey Fred -- just noticed that you have two commits for this PR. Per ODP policy, could you please squash these on your branch and then git push -f to condense into a single change?
> 
> Overall I think the code changes look good now

Yes, @xushengfei, ODP subscribed to a clean commit history policy with no automatic squashing of commits from PRs: https://github.com/OpenDevicePartnership/embassy-imxrt/blob/main/CONTRIBUTING.md#clean-commit-history.

Each commit in the history is expected to be grouped logically, functional, builds, and passes all workflows.

For this PR, it would be advisable to squash the second fixup commit into the first commit, because the first commit will introduce the u16 overflow bug. Thanks.

felipebalbi (changes requested):
Combine commits.

PR #358, opened 2025-03-27
Fix cannot set UART baudrate to those value other than 115200 issue

xushengfei:
In current Uart code, it only accepts 115200. If we set it to 3000000 or 4000000, the whole system will crash.
We need 3 changes to fix it:

Open clock setting to have user to set different type, for example: FFRO. Now it is hardcoded as SFRO.
Open source clock setting in Hz, in our PTL program, we need to set it as 48M, but it is hardcoded as 16M Hz.
There is a bug in current code to calculate brgval.
let brgval = (source_clock_hz / ((osrval + 1) * baudrate)) - 1;
if source_clock_hz is 48M and baudrate is 4M, osrval is 15, brgval is be -1, since the type is u32, it cannot be accept by rust, it will return error.
Need to use compare the number instead of compare it with FFFF.
I tested it in my device by setting 4M baudrarte, it can work as expected.

### DIFF
--- a/src/uart.rs
+++ b/src/uart.rs
@@ -11,6 +11,7 @@ use paste::paste;
 
 use crate::dma::channel::Channel;
 use crate::dma::transfer::Transfer;
+use crate::flexcomm::Clock;
 use crate::gpio::{AnyPin, GpioPin as Pin};
 use crate::interrupt::typelevel::Interrupt;
 use crate::iopctl::{DriveMode, DriveStrength, Inverter, IopctlPin, Pull, SlewRate};
@@ -74,6 +75,8 @@ pub struct Config {
     pub continuous_clock: Cc,
     /// Normal/ loopback mode
     pub loopback_mode: Loop,
+    /// Clock type
+    pub clock: Clock,
 }
 
 impl Default for Config {
@@ -89,6 +92,7 @@ impl Default for Config {
             sync_mode_master_select: Syncmst::Slave,
             continuous_clock: Cc::ClockOnCharacter,
             loopback_mode: Loop::Normal,
+            clock: crate::flexcomm::Clock::Sfro,
         }
     }
 }
@@ -295,9 +299,7 @@ impl<'a, M: Mode> Uart<'a, M> {
         cts: Option<Peri<'a, AnyPin>>,
         config: Config,
     ) -> Result<()> {
-        // TODO - clock integration
-        let clock = crate::flexcomm::Clock::Sfro;
-        T::enable(clock);
+        T::enable(config.clock);
         T::into_usart();
 
         let regs = T::info().regs;
@@ -320,23 +322,26 @@ impl<'a, M: Mode> Uart<'a, M> {
             regs.cfg().modify(|_, w| w.ctsen().enabled());
         }
 
-        Self::set_baudrate_inner::<T>(config.baudrate)?;
+        Self::set_baudrate_inner::<T>(config.baudrate, config.clock)?;
         Self::set_uart_config::<T>(config);
 
         Ok(())
     }
 
-    fn get_fc_freq() -> u32 {
-        // Todo: Make it generic for any clock
-        // Since the FC clock is hardcoded to Sfro, this freq is returned.
-        // sfro : 16MHz, // ffro: 48MHz
-        16_000_000
+    fn get_fc_freq(clock: Clock) -> Result<u32> {
+        match clock {
+            Clock::Sfro => Ok(16_000_000),
+            Clock::Ffro => Ok(48_000_000),
+            // We only support Sfro and Ffro now.
+            _ => Err(Error::InvalidArgument),
+        }
     }
 
-    fn set_baudrate_inner<T: Instance>(baudrate: u32) -> Result<()> {
-        let source_clock_hz = Self::get_fc_freq();
+    fn set_baudrate_inner<T: Instance>(baudrate: u32, clock: Clock) -> Result<()> {
+        // Get source clock frequency according to clock type.
+        let source_clock_hz = Self::get_fc_freq(clock)?;
 
-        if baudrate == 0 || source_clock_hz == 0 {
+        if baudrate == 0 {
             return Err(Error::InvalidArgument);
         }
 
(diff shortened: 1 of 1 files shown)


jerrysxie (approved):

src/uart.rs
```diff
@@ -371,24 +376,27 @@ impl<'a, M: Mode> Uart<'a, M> {
...
-                        // Calculate the difference between the
-                        // current baud rate and the desired baud rate
-                        diff = (candidate as i32 - baudrate as i32).unsigned_abs();
-
-                        // Check if the current calculated difference is the best so far
-                        if diff < best_diff {
-                            (diff, osrval, brgval)
-                        } else {
+                        let brgval = (source_clock_hz / ((osrval + 1) * baudrate)) - 1;
+                        // We know brgval will not be less than 0 now, it should have already been a valid u32 value now, then compare it with 65535.
```
felipebalbi, line 383: nit: This line needs to be broken after `value`. Max line width is set at 120 for this project.

src/uart.rs
```diff
@@ -371,24 +376,27 @@ impl<'a, M: Mode> Uart<'a, M> {
...
+                            // Calculate the difference between the
+                            // current baud rate and the desired baud rate
+                            let diff = (candidate as i32 - baudrate as i32).unsigned_abs();
+
+                            // Check if the current calculated difference is the best so far
+                            if diff < best_diff {
+                                (diff, osrval, brgval)
+                            } else {
+                                (best_diff, best_osr, best_brg)
+                            }
```
felipebalbi, line 399: It appears to me that all these changes to `set_baudrate_inner()` are unnecessary. The only required change to fix the bug is the change to get the clock frequency if another clock is selected as input, right? The calculations done here haven't really changed other than moving `brgval` compute later into the logic. Can you verify that this still works without most of the changes to `set_baudrate_inner()`?

```diff
diff --git a/src/uart.rs b/src/uart.rs
index 1014f94..6d81594 100644
--- a/src/uart.rs
+++ b/src/uart.rs
@@ -74,6 +74,8 @@ pub struct Config {
     pub continuous_clock: Cc,
     /// Normal/ loopback mode
     pub loopback_mode: Loop,
+    /// Clock type
+    pub clock: Clock,
 }

 impl Default for Config {
@@ -89,6 +91,7 @@ impl Default for Config {
             sync_mode_master_select: Syncmst::Slave,
             continuous_clock: Cc::ClockOnCharacter,
             loopback_mode: Loop::Normal,
+            clock: crate::flexcomm::Clock::Sfro,
         }
     }
 }
@@ -307,9 +310,7 @@ impl<'a, M: Mode> Uart<'a, M> {
         cts: Option<PeripheralRef<'_, AnyPin>>,
         config: Config,
     ) -> Result<()> {
-        // TODO - clock integration
-        let clock = crate::flexcomm::Clock::Sfro;
-        T::enable(clock);
+        T::enable(config.clock);
         T::into_usart();

         let regs = T::info().regs;
@@ -332,21 +333,23 @@ impl<'a, M: Mode> Uart<'a, M> {
             regs.cfg().modify(|_, w| w.ctsen().enabled());
         }

-        Self::set_baudrate_inner::<T>(config.baudrate)?;
+        Self::set_baudrate_inner::<T>(config.baudrate, config.clock)?;
         Self::set_uart_config::<T>(config);

         Ok(())
     }

-    fn get_fc_freq() -> u32 {
-        // Todo: Make it generic for any clock
-        // Since the FC clock is hardcoded to Sfro, this freq is returned.
-        // sfro : 16MHz, // ffro: 48MHz
-        16_000_000
+    fn get_fc_freq(clock: Clock) -> Result<u32> {
+        match clock {
+            Clock::Sfro => Ok(16_000_000),
+            Clock::Ffro => Ok(48_000_000),
+            // We only support Sfro and Ffro for now.
+            _ => Err(Error::InvalidArgument),
+        }
     }

-    fn set_baudrate_inner<T: Instance>(baudrate: u32) -> Result<()> {
-        let source_clock_hz = Self::get_fc_freq();
+    fn set_baudrate_inner<T: Instance>(baudrate: u32, clock: Clock) -> Result<()> {
+        let source_clock_hz = Self::get_fc_freq(clock);

         if baudrate == 0 || source_clock_hz == 0 {
             return Err(Error::InvalidArgument);
```

If this works fine then we should merge the minimal fix to avoid unexpected results. If the other changes were done for readability, we can do that as a separate PR. It's easier to review both changes if they are split into their logical units.

JamesHuard (approved):

src/uart.rs
xushengfei: Without this change in set_baudrate_inner, it will not work.

let brgval = (source_clock_hz / ((osrval + 1) * baudrate)) - 1;
if source_clock_hz is 48M, osrval is 15, baudrate is 4M, then (source_clock_hz / ((osrval + 1) * baudrate)) will be 0, brgval will be -1. but brgval is u32, -1 can be accepted in C language if it is u32, but in rust, it will cause a crash. 

That's why we need to compare wiht "if source_clock_hz < ((osrval + 1) * baudrate)" first to make sure brgval will not be less than 0.

felipebalbi: > Without this change in set_baudrate_inner, it will not work.
> 
> let brgval = (source_clock_hz / ((osrval + 1) * baudrate)) - 1; if source_clock_hz is 48M, osrval is 15, baudrate is 4M, then (source_clock_hz / ((osrval + 1) * baudrate)) will be 0, brgval will be -1. but brgval is u32, -1 can be accepted in C language if it is u32, but in rust, it will cause a crash.
> 
> That's why we need to compare wiht "if source_clock_hz < ((osrval + 1) * baudrate)" first to make sure brgval will not be less than 0.

good point, thanks.

felipebalbi (approved):

src/uart.rs
xushengfei: Thanks, update.

jerrysxie (approved):

felipebalbi (approved):

src/uart.rs
```diff
@@ -89,6 +92,7 @@ impl Default for Config {
             sync_mode_master_select: Syncmst::Slave,
             continuous_clock: Cc::ClockOnCharacter,
             loopback_mode: Loop::Normal,
+            clock: crate::flexcomm::Clock::Sfro,
```
huseinmsft, line 95: is there a reason to use sfro for default ? ffro covers most of our used baudrates and saves one line of config change on product code :)

xushengfei: The original setting in this file is sfro, the related settings in mis-766 should also base on this setting. this is to back compatible with mis-766. If mis-766 get latest code from embassy-imxrt, it still can work without any changes. :)

PR #361, opened 2025-03-28
Fixes for upstream part 2

felipebalbi:
There were a couple issues left related to two examples. CRC was using an exclusive borrow, that has been converted into a call to `reborrow()`. Timer was not wrapping Instances with `Peri`, now it is.


jerrysxie (approved):

JamesHuard (approved):
I think we're seeing some issues around gpio::Flex -- is this known?

felipebalbi:
> I think we're seeing some issues around gpio::Flex -- is this known?

Provided you run `cargo update`, there are no known issues. If you are seeing issues, then we need to figure out how to reproduce and fix them. Glad to chat privately if you prefer.

Issue #353, opened 2025-03-20
Slave/target I2C forever clock stretching after being addressed and first byte received


DM777-ms:
On back-to-back writes from the host controller/master to mcu i2c slave/target, the mcu i2c can be left in a forever clock stretched state after being addressed, then one byte received:

![Image](https://github.com/user-attachments/assets/8633d401-ff7e-444e-a6c1-264f23e273bc)

Details:
Repros when the time between the i2c writes are less than 80 us between the STOP of the previous write, and START of the next, failed write. 
400K Hz I2C clock
RT633 slave/target port =  FC4 / I2C4

Code under test is loosely based on the ODP HID Button Demo:
https://github.com/OpenDevicePartnership/hid-embedded-controller/blob/main/src/button/main.rs

Repros when two oem-specific info prints are in the Device and Host task loops that drive the slave hidi2c interface. With no info prints, bug does not repro:

```
const HID_ID0: hid::DeviceId = hid::DeviceId(0);

#[embassy_executor::task]
async fn kb_hidi2c_device_task(int_pin: Output<'static>) {
    define_static_buffer!(buffer, u8, [0; 64]);
    let buffer: OwnedRef<'static, u8> = buffer::get_mut().unwrap();

    static KB_DEVICE: OnceLock<KeyboardDevice> = OnceLock::new();
    let device = KB_DEVICE.get_or_init(|| KeyboardDevice::new(HID_ID0, buffer, int_pin));
    hid::register_device(device).await.unwrap();

    loop {
       oem_info!("HIDI2C Device");     // bug repros when this, and device task info print, are used
        device.process().await;
    }
}

#[embassy_executor::task]
async fn kb_hidi2c_host_task(mut i2c: I2cSlave<'static, Async>) {
    // apply I2cHostWrapper to i2c
    let i2c_hostw = subsystem_keyboard_hidi2c::I2cHostWrapper(i2c);

    define_static_buffer!(host_buffer, u8, [0; 128]);
    static HOST: OnceLock<Host<I2cHostWrapper<'static>>> = OnceLock::new();
    let host = HOST.get_or_init(|| Host::new(HID_ID0, i2c_hostw, host_buffer::get_mut().unwrap()));
    comms::register_endpoint(host, &host.tp).await.unwrap();

    loop {
       oem_info!("HIDI2C Host);      // bug repros when this, and device task info print, are used
        if let Err(e) = host.process().await {
[...]

PR #365, opened 2025-04-03
i2c slave: relax state machine restriction

jerrysxie:
This is harmless, if the slave is in the wrong state, it will just fail later, but this allows i2c slave state machine to be more robust if due to misuse of slave API, the slave is already past the addressed state.

--------------------------------------

This was exposed by an unexpectant use of the i2c slave API.

Normally we expect `respond_to_write()` to be invoked until it returns `Complete` to complete an I2C transaction, then `listen()` is invoked to wait for the next transaction.

However, in this case, due to a misunderstanding of the i2c slave API, `listen()` was invoked before `respond_to_write()` returns `Complete`. Causing the STOP for the previous i2c transaction to be interpreted as the STOP for the current ongoing transaction and thus thinking the current transaction is an 0-byte write. Completes the current transaction even though it is still in progress and stuck failing in `listen()` for all sequential transaction because it thinks it is in the addressed phase of the transaction even though it already move past the addressed phase into transmit or receive. The bug in the client code needs to be fixed: https://github.com/OpenDevicePartnership/embedded-services/issues/33. 

But there is also an opportunity to make the state machine more robust. If we relax addressed phase check, then this allows the state machine to move to the next state if it is already past the addressed phase. If the hardware is in a bad state already, it will just fail later at the end of the `listen()`.

### DIFF
--- a/src/i2c/slave.rs
+++ b/src/i2c/slave.rs
@@ -442,8 +442,9 @@ impl I2cSlave<'_, Async> {
         if i2c.stat().read().slvstate().is_slave_address() {
             i2c.slvctl().write(|w| w.slvcontinue().continue_());
         } else {
-            // If we are not addressed here, then we have issues.
-            return Err(TransferError::OtherBusError.into());
+            // If we are already past the addressed phase and in transmit or receive, that means we are already in the
+            // next state, most likely due to calling listen() before the previous transaction is completed and leading
+            // to state transition out of order. We can tolerate that, so we just move onto the next state.
         }
 
         // Poll for HW to transitioning from addressed to receive/transmit


RobertZ2011 (approved):

JamesHuard (dismissed):
Appreciate the description! Would it make sense to add any of this as a comment where the else case was?

jerrysxie:
> Appreciate the description! Would it make sense to add any of this as a comment where the else case was?

Good point, I will add it.

JamesHuard (approved):

DM777-ms:
Since this a temporary patch/workaround, will issue 353  remain open?
https://github.com/OpenDevicePartnership/embassy-imxrt/issues/353

DM777-ms (approved):

felipebalbi (approved):

PR #366, opened 2025-04-03
Move waker registrations PRIOR to condition checking

jamesmunns:
This is a defensive move to ensure that wakes are not ever lost, if the event we are waiting for happens AFTER checking, but BEFORE (re-)registering a waker, which can happen with interrupts.

In general, "Register, then check" is the pattern that should be used to avoid potential race conditions.

There are two cases I was not sure about:

* dma::channel seems to check twice, once before and once after registering, it is likely fine to just register and then check once.
* `rng` does a slightly more complicated check, potentially not unmasking interrupts. This might also be able to be simplified.

### DIFF
--- a/src/dma/channel.rs
+++ b/src/dma/channel.rs
@@ -87,6 +87,7 @@ impl<'d> Channel<'d> {
             let channel = self.info.ch_num;
 
             // Has the transfer already completed?
+            // TODO: Is this necessary? We could check once after registration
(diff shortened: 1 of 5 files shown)


felipebalbi (approved):
Looks good to me, thanks @jamesmunns

JamesHuard (approved):

Issue #303, opened 2025-02-10
I2C master should send a STOP after an address NACK


jerrysxie:
I2C spec is not very strict on how an I2C master should handle an address NACK. Both STOP and START are allowed.

However, it seems like some I2C slaves are not tolerant of the case where the master does not send a STOP.

PR #367, opened 2025-04-03
Add I2C DMA drop guards

jamesmunns:
This PR adds drop guards to disable the master/slave DMA on completion or cancellation of i2c futures.

I also added one refactor to avoid a duplicate check (and potential panicking branch, if the optimizer wasn't paying attention).

### DIFF
--- a/src/i2c/master.rs
+++ b/src/i2c/master.rs
@@ -479,17 +479,16 @@ impl<'a> I2cMaster<'a, Async> {
         let i2cregs = self.info.regs;
 
         // read of 0 size is not allowed according to i2c spec
-        if read.is_empty() {
+        //
+        // These are used in the DMA branch, but also checks implicitly
+        // if the slice is empty, so we do it here anyway
+        let Some((last_byte, dma_read)) = read.split_last_mut() else {
             return Err(TransferError::OtherBusError.into());
-        }
+        };
 
         self.start(address, true).await?;
 
         if self.dma_ch.is_some() {
-            // Do a DMA recv
-            // Read one byte less using DMA and then read the last byte manually
-            let (dma_read, last_byte) = read.split_at_mut(read.len() - 1);
-
             if !dma_read.is_empty() {
                 let transfer = dma::transfer::Transfer::new_read(
                     self.dma_ch.as_mut().unwrap(),
@@ -505,6 +504,12 @@ impl<'a> I2cMaster<'a, Async> {
                 // Additionally, at this point we know the slave has
                 // acknowledged the address.
                 i2cregs.mstctl().write(|w| w.mstdma().enabled());
+
+                // Use drop guard to ensure that DMA is disabled when we exit
+                // scope, successful or not.
+                let _dma_guard = OnDrop::new(|| {
+                    i2cregs.mstctl().write(|w| w.mstdma().disabled());
+                });
 
                 let res = select(
                     transfer,
@@ -532,8 +537,6 @@ impl<'a> I2cMaster<'a, Async> {
                     }),
                 )
                 .await;
-
-                i2cregs.mstctl().write(|w| w.mstdma().disabled());
 
                 if let Either::Second(e) = res {
                     e?;
@@ -568,7 +571,7 @@ impl<'a> I2cMaster<'a, Async> {
             .await?;
 
             // Read the last byte
-            last_byte[0] = i2cregs.mstdat().read().data().bits();
+            *last_byte = i2cregs.mstdat().read().data().bits();
         } else {
             // No DMA recv, we must manually recv one byte at a time.
             let read_len = read.len();
@@ -645,6 +648,12 @@ impl<'a> I2cMaster<'a, Async> {
             // acknowledged the address.
             i2cregs.mstctl().write(|w| w.mstdma().enabled());
 
+            // Use drop guard to ensure that DMA is disabled when we exit
+            // scope, successful or not.
+            let dma_guard = OnDrop::new(|| {
+                i2cregs.mstctl().write(|w| w.mstdma().disabled());
+            });
+
             let res = select(
                 transfer,
                 poll_fn(|cx| {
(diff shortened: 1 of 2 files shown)


felipebalbi (dismissed):

src/i2c/master.rs
```diff
@@ -506,6 +505,12 @@ impl<'a> I2cMaster<'a, Async> {
                 // acknowledged the address.
                 i2cregs.mstctl().write(|w| w.mstdma().enabled());
 
+                // Use drop guard to ensure that DMA is disabled when we exit
+                // scope, successful or not.
+                let _dma_guard = OnDrop::new(|| {
+                    i2cregs.mstctl().modify(|_r, w| w.mstdma().disabled());
```
jerrysxie, line 511: Instead of just disabling DMA on i2c DMA peripheral, we should also abort ongoing DMA if the DMA is still going.

We can call `abort()` to do that: https://github.com/OpenDevicePartnership/embassy-imxrt/blob/38d41e2c27f9363d607493e00a68aa3f5bb0822e/src/dma/channel.rs#L74.

And then we probably also want to make sure that a STOP will be sent to abort the ongoing I2C transaction.

jerrysxie (changes requested):

src/i2c/master.rs
jamesmunns: @jerrysxie I believe that the DMA transfer itself has a drop impl, meaning that on cancellation the transfer will be stopped and disabled.

https://github.com/OpenDevicePartnership/embassy-imxrt/blob/768a25b09a570226bfe06dce5e0e3b93ee980640/src/dma/transfer.rs#L191-L195

https://github.com/OpenDevicePartnership/embassy-imxrt/blob/768a25b09a570226bfe06dce5e0e3b93ee980640/src/dma/channel.rs#L73-L81

jerrysxie: Gotcha, that means sense. I always forgot to account for `drop()`: spooky action from a distance.

From a DMA perspective, it seems we are good there. It should be restored to a good state.

For I2C bus perspective, if we just stop transferring data in the middle of an I2C read, we will need to send a NACK + STOP to abort the transaction.

So we need to hit `i2cregs.mstctl().write(|w| w.mststop().set_bit());`
image

src/i2c/master.rs
```diff
@@ -651,7 +651,7 @@ impl<'a> I2cMaster<'a, Async> {
             // Use drop guard to ensure that DMA is disabled when we exit
             // scope, successful or not.
             let dma_guard = OnDrop::new(|| {
-                i2cregs.mstctl().write(|w| w.mstdma().disabled());
+                i2cregs.mstctl().modify(|_r, w| w.mstdma().disabled());
```
jerrysxie, line 654: For i2c write, we don't necessarily need to send a start but that would leave this master holding onto the bus. So we will want to hit `i2cregs.mstctl().write(|w| w.mststop().set_bit());` to abort the transaction.

src/i2c/slave.rs
```diff
@@ -531,6 +532,11 @@ impl I2cSlave<'_, Async> {
                 .unwrap()
                 .read_from_peripheral(i2c.slvdat().as_ptr() as *mut u8, buf, options);
 
+        // Hold guard to disable on cancellation or completion
+        let _dma_guard = OnDrop::new(|| {
+            i2c.slvctl().modify(|_r, w| w.slvdma().disabled());
```
jerrysxie, line 537: For I2C writes, if the slave aborts in the middle, we probably want to send a NACK to signal abort to the I2C master.
image

src/i2c/slave.rs
```diff
@@ -604,6 +610,11 @@ impl I2cSlave<'_, Async> {
                 .unwrap()
                 .write_to_peripheral(buf, i2c.slvdat().as_ptr() as *mut u8, options);
 
+        // Hold guard to disable on cancellation or completion
+        let _dma_guard = OnDrop::new(|| {
+            i2c.slvctl().modify(|_r, w| w.slvdma().disabled());
```
jerrysxie, line 615: For I2C reads, if i2c slave just stops sending data, then the I2C transaction would not completed if master is still waiting for more data. Not sure what is the best way to abort gracefully if this is dropped.

jerrysxie (changes requested):
How have we tested this? Using `with_timeout()` with a really short timeout to trigger the drop and make sure we can start a new I2C transaction with no errors?

jamesmunns:
Hey @jerrysxie, this was intended to be a non-invasive/minimal change to ensure that the "same behavior" occurs on completion or cancellation, would you prefer I address all your suggestions in this PR, or in a follow-up PR?

Thanks for all the feedback!

src/i2c/slave.rs
jamesmunns: I'm not sure either, this is the one that I haven't addressed.

src/i2c/master.rs
jamesmunns: Not so spooky :)

We do want to make sure that every future "cleans up" after itself, so this PR is aiming to bring that up to par for the I2C futures!

jamesmunns:
Addressed *most* of your comments @jerrysxie, should be good to review again.

Sending a stop/nak is a bit more complicated, because we can't do it when the engine is active. For this reason, I added the ability for this to be done by the interrupt handler automatically, since there is an "engine pending" interrupt we're already listening for,

jerrysxie (approved):

src/i2c/slave.rs
jerrysxie: Tracked in issue https://github.com/OpenDevicePartnership/embassy-imxrt/issues/375, so we don't forget.

jerrysxie: One option might be a send-more-data guard such until we get deselected or nacked, we keep sending another piece of data in the pending interrupt handler. But that can be another PR.

jerrysxie:
> Addressed _most_ of your comments @jerrysxie, should be good to review again.
> 
> Sending a stop/nak is a bit more complicated, because we can't do it when the engine is active. For this reason, I added the ability for this to be done by the interrupt handler automatically, since there is an "engine pending" interrupt we're already listening for,

This is a pretty elegant solution, James, love it! It also resolves issue #303 that I have been meaning to get around and fix for a while.

felipebalbi (approved):

PR #368, opened 2025-04-03
Add UART drop guards

jamesmunns:
Adds drop guards for UART for cancellation safety, similar to #367.

I also noticed that interrupts were never enabled when calling `Uart::new_async`, and that there were tabs in the macros, so I fixed those as well.

### DIFF
--- a/src/uart.rs
+++ b/src/uart.rs
@@ -770,6 +770,9 @@ impl<'a> Uart<'a, Async> {
         let tx = tx.into();
         let rx = rx.into();
 
+        T::Interrupt::unpend();
+        unsafe { T::Interrupt::enable() };
+
         let tx_dma = dma::Dma::reserve_channel(tx_dma);
         let rx_dma = dma::Dma::reserve_channel(rx_dma);
 
--- a/src/uart.rs
+++ b/src/uart.rs
@@ -5,6 +5,7 @@ use core::marker::PhantomData;
 use core::task::Poll;
 
 use embassy_futures::select::{select, Either};
+use embassy_hal_internal::drop::OnDrop;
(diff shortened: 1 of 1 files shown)


felipebalbi (approved):
Looks like we need to make sure `rustfmt` always uses spaces 🤦‍♂️

jamesmunns:
@felipebalbi in general, rustfmt gives up inside of most non-trivial macros. Since these macros were using `paste!`, rustfmt was probably never going to touch it.

I *think* rustfmt does convert tabs to spaces by default, but i'm not 100% on that.

jamesmunns:
(I think the only macros it DOESN'T give up are simple function-like macro invocations, like `println!()` or `log!()`, where it will format them as if they were a simple function call.

felipebalbi:
> @felipebalbi in general, rustfmt gives up inside of most non-trivial macros. Since these macros were using `paste!`, rustfmt was probably never going to touch it.
> 
> I _think_ rustfmt does convert tabs to spaces by default, but i'm not 100% on that.

in that case, I need to fix my `emacs` config LOL

src/uart.rs
```diff
@@ -770,6 +777,9 @@ impl<'a> Uart<'a, Async> {
         let tx = tx.into();
         let rx = rx.into();
 
+        T::Interrupt::unpend();
+        unsafe { T::Interrupt::enable() };
+
```
felipebalbi, line 782: Technically this is not needed, as `Uart` builds `UartRx` and `UartTx`, both of which will `unpend()` and `enabled()` the interrupt. Still, if we ever change the driver, maybe this is better.

RobertZ2011 (approved):

PR #369, opened 2025-04-03
Fix UART CTS 

huseinmsft:
set_uart_config() does a regs.cfg().write() which overwrites the regs.cfg().modify(|_, w| w.ctsen().enabled()); done earlier

### DIFF
--- a/src/uart.rs
+++ b/src/uart.rs
@@ -411,7 +411,7 @@ impl<'a, M: Mode> Uart<'a, M> {
     fn set_uart_config<T: Instance>(config: Config) {
         let regs = T::info().regs;
 
-        regs.cfg().write(|w| w.enable().disabled());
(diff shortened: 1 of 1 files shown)


JamesHuard (approved):

felipebalbi (changes requested):
Then set_uart_config() should just be converted to a `modify()` instead?

huseinmsft:
> Then set_uart_config() should just be converted to a `modify()` instead?

that should be ok too as the default value of that register is 0 anyways

felipebalbi (approved):
Awesome, thanks @huseinmsft. Nice catch :)

jerrysxie (dismissed):
Subtle bug, great catch! 💯

felipebalbi (dismissed):

felipebalbi (approved):

JamesHuard (approved):

PR #370, opened 2025-04-04
Ensure I2C blocking slave polling stops on deselect as well.

davidv1992:
This fixes i2c blocking slave functions not returning until the next operation starts.

### DIFF
--- a/src/i2c/slave.rs
+++ b/src/i2c/slave.rs
@@ -219,7 +219,7 @@ impl<'a> I2cSlave<'a, Blocking> {
     fn poll(&self) -> Result<()> {
         let i2c = self.info.regs;
 
-        while i2c.stat().read().slvpending().is_in_progress() {}
+        while i2c.stat().read().slvpending().is_in_progress() && i2c.stat().read().slvdesel().is_not_deselected() {}
 
         Ok(())
     }


davidv1992:
Hmm, not sure what clarification you are looking for, as the statement basically says it all, but to give concrete examples, it will not finish a respond_read for example until the master comes to the slave again for the next read or write operation, even if it has already sent a stop before that. Until that time, the future returned by that function remains pending.

I have added the additional text also to the commit message.

jerrysxie:
> Hmm, not sure what clarification you are looking for, as the statement basically says it all, but to give concrete examples, it will not finish a respond_read for example until the master comes to the slave again for the next read or write operation, even if it has already sent a stop before that. Until that time, the future returned by that function remains pending.
> 
> I have added the additional text also to the commit message.

@davidv1992 Thanks for finding this bug, I did not notice this because I was always testing with back-to-back transactions. My assumption was based on NXP's user manual where it claims that the pending bit will be set whenever the i2c peripheral needs SW servicing. It seems like in this case it will not be set in this case where a STOP is sent by the master, only the deselected bit is set.

jerrysxie (approved):

felipebalbi (approved):
Probably a good idea to have an example that shows how this would trigger.

PR #371, opened 2025-04-04
Added support for OOB and removed SRST added wait_for_plat_reset

philgweber:
After review of NXP part the proper sequence is first to wait for configuration to be completed and platform reset to be cleared before doing any other actions.

Added wait_for_plat_reset function that clients should wait for before doing further communication.

Added support for OOB mailbox and validated we can receive OOB packets. Future change coming to send OOB packets from MCU to host as well.

### DIFF
--- a/src/espi.rs
+++ b/src/espi.rs
@@ -161,7 +161,16 @@ pub enum PortConfig {
     PutPcMem32,
 
     /// Mailbox Split OOB
-    MailboxSplitOOB,
+    MailboxSplitOOB {
+        /// Port address to the host
+        addr: u16,
+
+        /// Offset into RAM space
+        offset: u16,
+
+        /// Length of the mailbox or mastering area per direction.
+        length: Len,
+    },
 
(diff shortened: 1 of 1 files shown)


src/espi.rs
```diff
@@ -576,6 +585,9 @@ impl<'d> Espi<'d> {
             } => {
                 self.mailbox(port, config.into(), direction, addr, offset, length);
             }
+            PortConfig::MailboxSplitOOB { addr, offset, length } => {
+                self.mailbox_oob(port, addr, offset, length);
```
felipebalbi, line 589: Use the more generic `mailbox()` method.

felipebalbi (changes requested):

src/espi.rs
philgweber: Changed to call mailbox with fixed direction rather than creating new function.

felipebalbi (approved):

jerrysxie (approved):

PR #372, opened 2025-04-04
Check Cargo.lock into source control and minor clean-up

jerrysxie:


felipebalbi (commented):
At least for the rolling workflow we want it to always take the latest such that we get to know when upstream introduces breaking changes. How about updating rolling.yml to make sure it runs cargo update prior to building?

felipebalbi (approved):

.vscode/settings.json
```diff
@@ -12,7 +12,8 @@
     "rust-analyzer.check.command": "clippy",
     "rust-analyzer.cargo.target": "thumbv8m.main-none-eabihf",
     "rust-analyzer.linkedProjects": [
-        "${workspaceFolder}/Cargo.toml",
+        //"${workspaceFolder}/Cargo.toml",
```
tullom, line 15: Should this be removed instead of commented out?

tullom (commented):
One minor nit comment, otherwise looks good,

RobertZ2011 (approved):

tullom (approved):

PR #374, opened 2025-04-04
Cargo vet init

### LINKED ISSUE #362: Init cargo vet audit

akshat2112:
Initializing `cargo vet` for this repository. This allows us to audit the third party dependencies being used in the code.

- As part of initialization, all current dependencies are under exemptions. These need to be audited as time permits.
- We import audits from Mozilla and Google. Dependencies that have been already audited by either of them are added to `imports.lock` file by `cargo vet`

A new workflow is also being added to run `cargo vet` on any new PR

Closes #362

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -6,6 +6,7 @@
 # - doc: checks that the code can be documented without errors
 # - hack: check combinations of feature flags
 # - msrv: check that the msrv specified in the crate is correct
+# - vet: checks for any new or updated unvetted dependencies
 permissions:
   contents: read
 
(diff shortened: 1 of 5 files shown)


.github/workflows/check.yml
```diff
@@ -312,3 +313,28 @@ jobs:
...
+      run: echo "${{ runner.tool_cache }}/cargo-vet/bin" >> $GITHUB_PATH
+
+    - name: Ensure that the tool cache is populated with the cargo-vet binary
+      run: cargo install --root ${{ runner.tool_cache }}/cargo-vet --version ${{ env.CARGO_VET_VERSION }} cargo-vet
+
+    - name: Ensure Cargo.lock is up-to-date
+      run: cargo generate-lockfile
+
+    - name: Invoke cargo-vet
+      run: cargo vet --locked
```
felipebalbi, line 340: Do we need this to run on all commits, like the other workflows in this file?

jerrysxie: Technically, we don't someone to add an unvetted dependency in one commit and remove it from another, right?

.github/workflows/check.yml
```diff
@@ -312,3 +313,28 @@ jobs:
         run: |
           cargo check -F mimxrt685s,rt,defmt,time,time-driver,unstable-pac
           cargo check -F mimxrt633s,rt,defmt,time,time-driver,unstable-pac
+  
+  vet:
```
jerrysxie, line 317: Can you outline what would we have to do after each `cargo update`?

1. Audit any new dependency
2. Update config.toml
3. Re-generate imports.lock

.github/workflows/check.yml
akshat2112: Good point. I did think about it but thought we only care about the final state of the repo and the dependencies being used then. Adding and removing a dependency is one attack vector that was missed. Will update

.github/workflows/check.yml
akshat2112: Added README file describing the process

jerrysxie (approved):
This changes should be made in the template repo to make sure all future repo have `cargo vet` as well:  https://github.com/OpenDevicePartnership/embedded-rust-template

felipebalbi (approved):

PR #376, opened 2025-04-08
Added support to trigger OOB read to host

philgweber:
The eSPI service needs to populate data into the correct eSPI memory area and specify the port and length of data to send.

Calling oob_read_start will send VWire event to the host to cause it to trigger a read of the data that is already loaded. After the host has finished reading the data it will send back an INTRD event on the corresponding port to indicate the read has completed.

The eSPI service must not trigger a new read until the previous read has completed.

### DIFF
--- a/src/espi.rs
+++ b/src/espi.rs
@@ -814,6 +814,25 @@ impl<'d> Espi<'d> {
         self.block_for_vwire_done();
     }
 
+    /// Trigger OOB read from host
+    /// When this completes it will send a INTWR event on OOB port
+    /// Maximum len is 64
+    pub fn oob_read_start(&mut self, port: usize, length: u8) {
+        // SAFETY: length can be any value in bits 6:0, bit 7 is reserved 0
+        self.info
+            .regs
+            .port(port)
+            .omflen()
+            .write(|w| unsafe { w.len().bits(length) });
+        // SAFETY: 0x2 = Completed by MCU in SSCTL register
+        // PAC definition is being updated to expose enum of valid values
+        self.info
+            .regs
+            .port(port)
+            .irulestat()
+            .modify(|_, w| unsafe { w.sstcl().bits(0x2) });
+    }
+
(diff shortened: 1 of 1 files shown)


src/espi.rs
```diff
@@ -814,6 +814,18 @@ impl<'d> Espi<'d> {
         self.block_for_vwire_done();
     }
 
+    /// Trigger OOB read from host
+    /// When this completes it will send a INTWR event on OOB port
+    /// Maximum len is 64
+    pub fn oob_read_start(&mut self, port: usize, len: u16) {
+        self.info.regs.port(port).addr().write(|w| unsafe { w.off().bits(len) });
```
tullom, line 821: // # SAFETY comment missing

src/espi.rs
```diff
@@ -814,6 +814,18 @@ impl<'d> Espi<'d> {
...
+        self.info
+            .regs
+            .port(port)
+            .irulestat()
+            .modify(|_, w| unsafe { w.sstcl().bits(0x2) });
```
tullom, line 826: // # SAFETY comment missing

felipebalbi (approved):

src/espi.rs
philgweber: Changed to use omflen which is correct mirror of this register and added SAFETY comment information.

src/espi.rs
philgweber: Added safety comment and started change for PAC to define enumeration for SSTCL so we don't need to do unsafe operation here in the future.

jerrysxie (approved):
Just for my understanding, when we get the event back, it will be just like peripheral channel: https://github.com/OpenDevicePartnership/embassy-imxrt/blob/ef4049531daa56d25307dd8ba276f01dc3c78c54/src/espi.rs#L647C1-L648C1, with direction, base, and offset?

felipebalbi (approved):

PR #389, opened 2025-04-11
Reworked blocking i2c slave interface

davidv1992:
Main motivators are:
 - More consistent interface that is better misuse resistant.
 - Easier to extend with support for Automatic Ack/Nack of addresses in async variant.
 - Better support for cancelation and capable of better surfacing what the hardware can actually do.

The code changes also fixes some issues in 10-bit addressing. It should now match the behaviour specified in UM10204

### DIFF
--- a/examples/rt685s-evk/src/bin/i2c-loopback-async-test.rs
+++ b/examples/rt685s-evk/src/bin/i2c-loopback-async-test.rs
@@ -1,0 +1,1870 @@
+#![no_std]
+#![no_main]
+
+extern crate embassy_imxrt_examples;
+
+use cortex_m::asm::delay;
+use defmt::info;
+use embassy_executor::Spawner;
+use embassy_imxrt::gpio::{DriveMode, DriveStrength, Flex, GpioPin, SenseEnabled, SlewRate};
+use embassy_imxrt::i2c::slave::{
+    Address, AsyncI2cSlave, ReadResult, Transaction, TransactionExpectRead, TransactionExpectWrite, WriteResult,
+};
+use embassy_imxrt::peripherals::{DMA0_CH4, FLEXCOMM2, PIO0_17, PIO0_18, PIO0_29, PIO0_30};
+use embassy_imxrt::{bind_interrupts, i2c, peripherals, Peri};
+use embassy_sync::blocking_mutex::raw::CriticalSectionRawMutex;
+use embassy_sync::semaphore::{GreedySemaphore, Semaphore as _};
+
+bind_interrupts!(struct Irqs {
+    FLEXCOMM2 => i2c::InterruptHandler<peripherals::FLEXCOMM2>;
+});
+
+type Semaphore = GreedySemaphore<CriticalSectionRawMutex>;
+
+struct I2CTester {
+    scl: Flex<'static, SenseEnabled>,
+    sda: Flex<'static, SenseEnabled>,
+}
+
+impl I2CTester {
+    fn new(scl: Peri<'static, impl GpioPin>, sda: Peri<'static, impl GpioPin>) -> Self {
+        let mut scl = Flex::<'_, SenseEnabled>::new(scl);
+        let mut sda = Flex::<'_, SenseEnabled>::new(sda);
+
+        scl.set_high();
+        scl.set_as_output(DriveMode::OpenDrain, DriveStrength::Normal, SlewRate::Standard);
+
+        sda.set_high();
+        sda.set_as_output(DriveMode::OpenDrain, DriveStrength::Normal, SlewRate::Standard);
+
+        I2CTester { scl, sda }
+    }
+
+    async fn start(&mut self) {
+        self.sda.set_high();
+        delay(1500);
+
+        self.scl.set_high();
+        self.scl.wait_for_high().await;
+        delay(1500);
+
+        self.sda.set_low();
+        delay(1500);
+
+        self.scl.set_low();
+        delay(1500);
+    }
+
+    async fn stop(&mut self) {
+        self.sda.set_low();
+        delay(1500);
+
+        self.scl.set_high();
+        self.scl.wait_for_high().await;
+        delay(1500);
+
+        self.sda.set_high();
+        delay(1500);
+    }
+
+    async fn i2c_byte(&mut self, mut send: u8, ack: bool) -> (u8, bool) {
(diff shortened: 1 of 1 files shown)


davidv1992:
This is only the rewrite of the blocking interface. If this design is considered reasonable I can extend it to the async side and actually do the support around automatic address ack/nack. The choice for separate types for the blocking and async slave interfaces is that both will need different code to run on drop, which cannot be done another way because of restrictions on Drop implementations.

src/i2c/slave.rs
```diff
@@ -203,141 +215,181 @@ impl<'a, M: Mode> I2cSlave<'a, M> {
...
+                } else {
+                    // Read transaction
+                    break Ok(Transaction::Read {
+                        handler: BlockingI2cSlaveRead {
+                            info: self.base.info,
+                            should_ack_addr: true,
+                            _phantom: PhantomData,
+                        },
+                    });
+                }
```
tdittr, line 347: I feel like this is too deeply nested, could we write this as a match, or split it out into separate functions (for 7-/10/bit addresses)?

src/i2c/slave.rs
tdittr, line 1: In general I think we should improve the granularity of the returned errors to make debugging easier.

tdittr (commented):
A few thoughts for now, going to finish the review next week.

src/i2c/slave.rs
```diff
@@ -350,40 +402,89 @@ impl I2cSlave<'_, Blocking> {
...
             xfer_count += 1;
         }
 
         let stat = i2c.stat().read();
         if stat.slvdesel().is_deselected() {
-            // Clear the deselect bit
-            i2c.stat().write(|w| w.slvdesel().deselected());
-            return Ok(Response::Complete(xfer_count));
+            // inhibit drop
+            core::mem::forget(self);
```
tdittr, line 416: Nit-pick: I would replace with a destructuring assignment.
```suggestion
            let BlockingI2cSlaveWrite{..} = self;
```

Or alternatively have a `Self::defuse` method.

src/i2c/slave.rs
```diff
@@ -409,26 +510,141 @@ impl I2cSlave<'_, Blocking> {
...
+impl Drop for BlockingI2cSlaveRead<'_> {
+    fn drop(&mut self) {
+        let i2c = self.info.regs;
+        if self.should_ack_addr {
+            i2c.slvctl().modify(|_, w| w.slvnack().nack());
+        } else {
+            // complete the read with all-ones overrun
+            loop {
+                // Block until something happens
+                blocking_poll(self.info);
```
tdittr, line 561: I think we should move handling the fill character into the interrupt handler, blocking in `drop` seems a bin unexpected to me.

examples/rt685s-evk/src/bin/i2c-slave.rs
tdittr, line 1: I like this example for showing a real-world use-case :+1:

tdittr (commented):
I like the new API, however it does feel a bit rough around the edges.

Should we maybe implement the `async` version and improve the documentation before merging?

jerrysxie (commented):
@davidv1992 Still going through the review, just one question about the overall goals and next steps:

1. For all the ODP I2C slave HALs, the plan is to have a HW agnostic traits interface to be implemented by both `embassy-imxrt` and `embassy-npcx`, is this a step toward that direction? The I2C traits are eventually to be upstreamed to embedded-hal, but meanwhile we should probably have it as a independent crate in ODP until the upstreaming can happen.
2. We do have some clients that relies on this i2c slave interface already. Merging this will require a bit coordination.

davidv1992:
@jerrysxie Yes, this is aimed to be a step in that direction. The exact shape of a hardware agnostic trait is still in the design phase, but something along these lines is what we are trending towards. The main pain points right now (though those are less relevant for the two crates you mention) is that some hardware has quite limited i2c slave support.

As for this requiring some coordination to merge, that is not entirely unexpected. My main goal at this stage is to gather feedback on the overall shape of the design in preparation for doing the same reworking of the interface in the async case. So far, reactions seem positive, but if on your end you feel different about it it would be nice to catch that before I have sunk a bunch of work into the async version, as it's likely going to be somewhat more complicated to implement.

src/i2c/slave.rs
davidv1992: For async this is indeed the better approach, however the blocking interface doesn't have an interrupt handler.

src/i2c/slave.rs
davidv1992: Done.

src/i2c/slave.rs
davidv1992: Fixed this in a different way, the drops are now short even for the blocking implementation.

src/i2c/slave.rs
davidv1992: All these errors are essentially identical (bus is in a strange state), so there isn't really anything to separate them.

jeffglaum:
@JamesHuard , @felipebalbi , @RobertZ2011 can you guys please review and provide feedback?  Thanks.

davidv1992:
@jeffglaum You have assigned this one back to me, but it is unclear what is required from me at this point. Could you clarify that?

src/i2c/slave.rs
```diff
@@ -201,143 +440,312 @@ impl<'a, M: Mode> I2cSlave<'a, M> {
...
             _flexcomm: flexcomm,
             _phantom: PhantomData,
-            dma_ch,
             ten_bit_info,
+            ten_bit_read_possible: false,
         })
     }
 }
 
-impl<'a> I2cSlave<'a, Blocking> {
```
felipebalbi, line 212: Just curious: why remove the `Mode` type parameter here?

felipebalbi (approved):
I didn't attempt to run the examples yet, but the code looks easy enough to read. Since you're using auto-ack, did you manage to quantify any changes to bus utilization?

jeffglaum:
> @jeffglaum You have assigned this one back to me, but it is unclear what is required from me at this point. Could you clarify that?

Just assigning to you, to complete the PR (looks like there are still open discussion points which are blocking the merge).  Thanks.

jeffglaum:
@davidv1992 are you intending to complete this PR (looks like it had approvals) or to abandon?  Thanks.

PR #392, opened 2025-04-14
Fixed an error in seconds to datetime conversion function as well as addition of get/set functions using seconds

Sebastian61:
Error in conversion using `convert_secs_to_datetime` function. Actual output is consistently off by 1 day due to erroneous initialization of `day` variable to `0` instead of `1`.

image

Added functions to get and set RTC register using seconds directly for more flexibility.

### DIFF
--- a/src/time_driver.rs
+++ b/src/time_driver.rs
@@ -470,6 +470,20 @@ impl<'r> RtcDatetime<'r> {
         Ok(())
     }
 
+    pub fn set_datetime_in_secs(&self, secs: u32) -> Result<(), Error> {
+        let r = rtc();
+        let datetime = &self.convert_secs_to_datetime(secs);
+        // SAFETY: Clear RTC_EN bit before setting time to handle race condition
+        //         when the count is in middle of a transition
+        //         There is 21 mS inacurracy in the time set
+        //         Todo: https://github.com/OpenDevicePartnership/embassy-imxrt/issues/121
+        r.ctrl().modify(|_r, w| w.rtc_en().disable());
+        self.is_valid_datetime(datetime)?;
+        r.count().write(|w| unsafe { w.bits(secs) });
+        r.ctrl().modify(|_r, w| w.rtc_en().enable());
+        Ok(())
+    }
+
     /// Get the datetime.
     pub fn get_datetime(&self) -> (Datetime, Result<(), Error>) {
         let r = rtc();
@@ -490,6 +504,29 @@ impl<'r> RtcDatetime<'r> {
         let res = self.is_valid_datetime(&datetime);
         {
             (datetime, res)
+        }
+    }
+
+    /// Get the datetime as UTC seconds
+    pub fn get_datetime_as_secs(&self) -> (u32, Result<(), Error>) {
+        let r = rtc();
+        //  If RTC is not enabled return error
+        if r.ctrl().read().rtc_en().bit_is_clear() {
+            return (0, Err(Error::RTCNotEnabled));
+        }
+        let secs: u32;
+        loop {
+            let secs1 = r.count().read().bits();
+            let secs2 = r.count().read().bits();
+            if secs1 == secs2 {
+                secs = secs1;
+                break;
+            }
+        }
+        let datetime = self.convert_secs_to_datetime(secs);
+        let res = self.is_valid_datetime(&datetime);
+        {
+            (secs, res)
         }
     }
 }
--- a/src/time_driver.rs
+++ b/src/time_driver.rs
@@ -405,7 +405,7 @@ impl<'r> RtcDatetime<'r> {
 
         let mut year = 1970;
         let mut month = 1;
-        let mut day = 0;
+        let mut day = 1;
 
         // Calculate year
         while days >= 365 {


src/time_driver.rs
```diff
@@ -470,6 +470,20 @@ impl<'r> RtcDatetime<'r> {
         Ok(())
     }
 
+    pub fn set_datetime_in_secs(&self, secs: u32) -> Result<(), Error> {
```
jerrysxie, line 473: This function `set_datetime_in_secs()` is almost the same as `set_datetime()` other than the unit, can we just have this function call `set_datetime()` instead?

src/time_driver.rs
```diff
@@ -492,6 +506,29 @@ impl<'r> RtcDatetime<'r> {
             (datetime, res)
         }
     }
+
+    /// Get the datetime as UTC seconds
+    pub fn get_datetime_as_secs(&self) -> (u32, Result<(), Error>) {
```
jerrysxie, line 511: This function `get_datetime_as_secs()` is almost the same as `get_datetime()` other than the unit, can we just `get_datetime()` call `get_datetime_as_secs()` and then convert second to datetime instead?

jerrysxie (dismissed):
@Sebastian61 We would like to expose a RTC HW agnostic traits to the upper layer application code. Do you know any generic RTC traits that this NXP RTC driver can implement? Or do we need to create one ourselves for now?

felipebalbi:
> @Sebastian61 We would like to expose a RTC HW agnostic traits to the upper layer application code. Do you know any generic RTC traits that this NXP RTC driver can implement? Or do we need to create one ourselves for now?

you could impl `[Datelike](https://docs.rs/chrono/0.4.40/chrono/trait.Datelike.html
)` and `[Timelike](https://docs.rs/chrono/0.4.40/chrono/trait.Timelike.html
)` from `chrono`.

At least 2 embassy HALs already implement it:

1. https://github.com/embassy-rs/embassy/blob/main/embassy-rp/src/rtc/datetime_chrono.rs
2. https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/rtc/datetime.rs

src/time_driver.rs
```diff
@@ -492,6 +506,29 @@ impl<'r> RtcDatetime<'r> {
...
+            if secs1 == secs2 {
+                secs = secs1;
+                break;
+            }
+        }
+        let datetime = self.convert_secs_to_datetime(secs);
+        let res = self.is_valid_datetime(&datetime);
+        {
+            (secs, res)
+        }
```
felipebalbi, line 530: why the block?

src/time_driver.rs
```diff
@@ -492,6 +506,29 @@ impl<'r> RtcDatetime<'r> {
...
+        let secs: u32;
+        loop {
+            let secs1 = r.count().read().bits();
+            let secs2 = r.count().read().bits();
+            if secs1 == secs2 {
+                secs = secs1;
+                break;
+            }
+        }
```
felipebalbi, line 525: Do you mind explaining what are you attempting here? Why do you need to potentially delay returning the value from `count` register?

src/time_driver.rs
```diff
@@ -470,6 +470,20 @@ impl<'r> RtcDatetime<'r> {
...
+        let r = rtc();
+        let datetime = &self.convert_secs_to_datetime(secs);
+        // SAFETY: Clear RTC_EN bit before setting time to handle race condition
+        //         when the count is in middle of a transition
+        //         There is 21 mS inacurracy in the time set
+        //         Todo: https://github.com/OpenDevicePartnership/embassy-imxrt/issues/121
+        r.ctrl().modify(|_r, w| w.rtc_en().disable());
+        self.is_valid_datetime(datetime)?;
+        r.count().write(|w| unsafe { w.bits(secs) });
+        r.ctrl().modify(|_r, w| w.rtc_en().enable());
```
felipebalbi, line 483: don't you need a critical section here? IIRC, `modify()` is not atomic.

src/time_driver.rs
Sebastian61: The `get_datetime()` function also takes the RTC register count as seconds before conversion into `DateTime` so I added the `get_datetime_as_secs()` function to cut out the needless conversion.

src/time_driver.rs
Sebastian61: Hi @felipebalbi I originally just took the `get_datetime()` and `set_datetime()` functions and edited them to take `secs: u32` instead but I've rewritten them now to be a bit cleaner.

src/time_driver.rs
Sebastian61: Hi @felipebalbi I originally just took the `get_datetime()` and `set_datetime()` functions and edited them to take `secs: u32` instead but I've rewritten them now to be a bit cleaner.

src/time_driver.rs
Sebastian61: Hi @felipebalbi I originally just took the `get_datetime()` and `set_datetime()` functions and edited them to take `secs: u32` instead but I've rewritten them now to be a bit cleaner.

src/time_driver.rs
jerrysxie: Sorry, I meant instead of duplicating the logic. We can just extract the common logic into a common function to call into. It seems like `set_datetime_in_secs()` can just call into `set_datetime()` and then do the additional conversion and vice versa.

src/time_driver.rs
jerrysxie: Same comment about duplicating logic, see https://github.com/OpenDevicePartnership/embassy-imxrt/pull/392#discussion_r2051501987.

src/time_driver.rs
Sebastian61: Hi I agree with extracting the common logic but surely that would mean ‘set_datetime()’ calling set_datetime_in_secs()’ as the common logic is writing into the RTC register which ‘set_datetime()’ has extra prerequisite steps?

jerrysxie: Sorry, I didn't notice there is a new commit. Let me take a look.

jerrysxie (changes requested):
Please rebase instead using merge, and use squash to get rid of the merge commit in history. We strive for a clean history : https://github.com/OpenDevicePartnership/embassy-imxrt/blob/main/CONTRIBUTING.md#clean-commit-history.

Sebastian61:
Hi @jerrysxie I've rebased this branch and addressed the unsafe write into the RTC count register.

huseinmsft (approved):

src/time_driver.rs
```diff
@@ -470,6 +470,15 @@ impl<'r> RtcDatetime<'r> {
         Ok(())
     }
 
+    ///Set the datetime in seconds
+    pub fn set_datetime_in_secs(&self, secs: u32) -> Result<(), Error> {
```
jerrysxie, line 471: Does this still need to return a result? It cannot fail anymore.

jerrysxie (changes requested):

src/time_driver.rs
```diff
@@ -481,25 +478,19 @@ impl<'r> RtcDatetime<'r> {
 
     /// Get the datetime.
     pub fn get_datetime(&self) -> (Datetime, Result<(), Error>) {
```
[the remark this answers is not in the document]
Sebastian61: I did notice that when making changes, but I didn't change it due to worries it might break some dependencies. I'll update it to return `Result<Datetime, Error>`.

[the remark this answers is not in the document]
Sebastian61: Updated to only return `Result<Datetime, Error>`

jerrysxie (approved):

PR #398, opened 2025-04-15
Make I2C duty cycle configurable

### LINKED ISSUE #267: i2c master SCL high and low time needs to be tunable
i2c master SCL high and low time needs to be tunable to allow OEM to tune it on a per layout basis.

https://github.com/pop-project/embassy-imxrt/blob/2322cefbcc108705462116b0dacb19df1f632d68/src/i2c/master.rs#L88

williampMSFT:
This change enables configuring the duty cycle for the I2C bus by specifying the number of clock cycles that the clock line will be asserted high/low by the I2C master device

Closes #267

### DIFF
--- a/examples/rt685s-evk/src/bin/i2c-loopback-async-10bit.rs
+++ b/examples/rt685s-evk/src/bin/i2c-loopback-async-10bit.rs
@@ -5,7 +5,7 @@ extern crate embassy_imxrt_examples;
 
 use defmt::info;
 use embassy_executor::Spawner;
-use embassy_imxrt::i2c::master::{I2cMaster, Speed};
+use embassy_imxrt::i2c::master::I2cMaster;
 use embassy_imxrt::i2c::slave::{Address, Command, I2cSlave, Response};
 use embassy_imxrt::i2c::{self, Async};
 use embassy_imxrt::{bind_interrupts, peripherals};
@@ -138,7 +138,7 @@ async fn main(spawner: Spawner) {
 
     let slave = I2cSlave::new_async(p.FLEXCOMM2, p.PIO0_18, p.PIO0_17, Irqs, SLAVE_ADDR.unwrap(), p.DMA0_CH4).unwrap();
 
-    let master = I2cMaster::new_async(p.FLEXCOMM4, p.PIO0_29, p.PIO0_30, Irqs, Speed::Standard, p.DMA0_CH9).unwrap();
+    let master = I2cMaster::new_async(p.FLEXCOMM4, p.PIO0_29, p.PIO0_30, Irqs, Default::default(), p.DMA0_CH9).unwrap();
 
     spawner.must_spawn(master_service(master));
     spawner.must_spawn(slave_service(slave));
--- a/examples/rt685s-evk/src/bin/i2c-loopback-async.rs
+++ b/examples/rt685s-evk/src/bin/i2c-loopback-async.rs
@@ -5,7 +5,7 @@ extern crate embassy_imxrt_examples;
 
 use defmt::info;
 use embassy_executor::Spawner;
-use embassy_imxrt::i2c::master::{I2cMaster, Speed};
+use embassy_imxrt::i2c::master::{DutyCycle, I2cMaster};
 use embassy_imxrt::i2c::slave::{Address, Command, I2cSlave, Response};
 use embassy_imxrt::i2c::{self, Async};
 use embassy_imxrt::{bind_interrupts, peripherals};
@@ -108,7 +108,11 @@ async fn main(spawner: Spawner) {
 
     let slave = I2cSlave::new_async(p.FLEXCOMM2, p.PIO0_18, p.PIO0_17, Irqs, SLAVE_ADDR.unwrap(), p.DMA0_CH4).unwrap();
 
-    let master = I2cMaster::new_async(p.FLEXCOMM4, p.PIO0_29, p.PIO0_30, Irqs, Speed::Standard, p.DMA0_CH9).unwrap();
+    let config = i2c::master::Config {
+        speed: i2c::master::Speed::Fast,
+        duty_cycle: DutyCycle::new(50).unwrap(),
+    };
+    let master = I2cMaster::new_async(p.FLEXCOMM4, p.PIO0_29, p.PIO0_30, Irqs, config, p.DMA0_CH9).unwrap();
 
     spawner.must_spawn(master_service(master));
     spawner.must_spawn(slave_service(slave));
--- a/src/i2c/master.rs
+++ b/src/i2c/master.rs
@@ -92,9 +92,7 @@ impl<'a, M: Mode> I2cMaster<'a, M> {
             // SAFETY: only unsafe due to .bits usage
             unsafe { w.mstsclhigh().bits(0).mstscllow().bits(1) });
 
-        regs.intenset().write(|w|
-                // SAFETY: only unsafe due to .bits usage
-                unsafe { w.bits(0) });
+        regs.intenset().reset();
 
         regs.cfg().write(|w| w.msten().set_bit());
 
--- a/src/i2c/master.rs
+++ b/src/i2c/master.rs
@@ -6,6 +6,7 @@ use core::task::Poll;
 
 use embassy_futures::select::{select, Either};
 use embassy_hal_internal::drop::OnDrop;
+use itertools::Itertools;
 
 use super::{
     force_clear_remediation, wait_remediation_complete, Async, Blocking, Error, Info, Instance, InterruptHandler,
@@ -14,9 +15,11 @@ use super::{
 };
 use crate::flexcomm::FlexcommRef;
 use crate::interrupt::typelevel::Interrupt;
+use crate::pac::i2c0::msttime::{Mstsclhigh, Mstscllow};
 use crate::{dma, interrupt, Peri};
 
 /// Bus speed (nominal SCL, no clock stretching)
+#[derive(Clone, Copy)]
 pub enum Speed {
     /// 100 kbit/s
     Standard,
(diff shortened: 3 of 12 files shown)


examples/rt685s-evk/src/bin/i2c-loopback-async-10bit.rs
```diff
@@ -138,7 +138,17 @@ async fn main(spawner: Spawner) {
...
 
-    let master = I2cMaster::new_async(p.FLEXCOMM4, p.PIO0_29, p.PIO0_30, Irqs, Speed::Standard, p.DMA0_CH9).unwrap();
+    let master = I2cMaster::new_async(
+        p.FLEXCOMM4,
+        p.PIO0_29,
+        p.PIO0_30,
+        Irqs,
+        Speed::Standard,
+        i2c::master::Mstscllow::Clocks3,
+        i2c::master::Mstsclhigh::Clocks2,
```
felipebalbi, line 148: This Low/High configuration is very coupled to the underlying MCU. Instead of exposing that, how about letting the user tell you the duty cycle they want and the driver computes Mstscllow and Mstsclhigh internally?

src/i2c/master.rs
```diff
@@ -31,6 +31,22 @@ pub enum Speed {
     High,
 }
 
+/// Master SCL Low time. Specifies the minimum low time in clock cycles that will be asserted by this master on SCL.
+#[cfg(feature = "mimxrt685s")]
+pub type Mstscllow = mimxrt685s_pac::i2c0::msttime::Mstscllow;
```
felipebalbi, line 36: `pub use mimxrt685s_pac::i2c0::msttime::Mstscllow;` will work the same way, I think

src/i2c/master.rs
```diff
@@ -31,6 +31,22 @@ pub enum Speed {
     High,
 }
 
+/// Master SCL Low time. Specifies the minimum low time in clock cycles that will be asserted by this master on SCL.
+#[cfg(feature = "mimxrt685s")]
```
felipebalbi, line 35: why only 685?

src/i2c/master.rs
```diff
@@ -127,9 +148,19 @@ impl<'a> I2cMaster<'a, Blocking> {
...
         force_clear_remediation(&T::info());
-        Ok(Self::new_inner::<T>(fc, scl, sda, speed, None)?)
+        Ok(Self::new_inner::<T>(
+            fc,
+            scl,
+            sda,
+            speed,
+            scl_low_clocks,
+            scl_high_clocks,
+            None,
```
felipebalbi, line 162: I think we've delayed it long enough; These arguments should be part of a `struct Config` to make it easier for the user. Look at other HALs in upstream embassy and the eSPI driver for example

src/i2c/master.rs
williampMSFT: I'm not sure I understand the question - 633 has a similar thing for its enum type on line 44 if that's what you're asking?

src/i2c/master.rs
williampMSFT: Sure, moved the 'C-enum' args to a struct. I left the 'which pins to bind to' piece out of the config because I don't think there are sane defaults and they're templated so we'd have to template the config struct, which I think could get messy? Let me know if you think those should move into the struct too and we can discuss

examples/rt685s-evk/src/bin/i2c-loopback-async-10bit.rs
felipebalbi: you choose the closest match and produce a `u32` that tells the user the selected duty cycle. That is:

```rust
fn set_duty(duty: usize) -> usize {
    let selected = self.find_closest_duty_cycle_to(duty);
    selected
}
```

Or something like that :)

src/i2c/mod.rs
```diff
@@ -21,21 +21,6 @@ pub mod slave;
...
-
-    /// 400 kbit/s
-    Fast,
-
-    /// 1 Mbit/s
-    FastPlus,
-
-    /// 3.4Mbit/s only available for slave devices
-    High,
-}
```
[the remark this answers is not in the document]
felipebalbi: I see, thanks for clarifying. You probably want this hunk as a separate commit (same PR), then.

src/i2c/master.rs
williampMSFT: oh, cool, thanks! that seems way cleaner, will do that instead

src/i2c/mod.rs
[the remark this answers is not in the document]
williampMSFT: Sure, broke out into a separate commit

examples/rt685s-evk/src/bin/i2c-loopback-async-10bit.rs
williampMSFT: I understand how one would write an implementation that tries to infer number of clocks given a percentage, but I think I'm unclear about how that interface would be used by clients on a couple points:

1) if the user wants 7 up/7 down on one master and 2 up/2 down on another master, how would they express that?  A percentage seems insufficient for this purpose, but if we are willing to say "nobody will ever want that", then this isn't an issue; I don't know if we're willing to say that, though.

2) Assuming the user will only ever want the shortest possible number of clocks for a given duty cycle percentage, there are only 45 discrete ratios possible (not 64 because duplicates), and the spacing in percentage points a them is nonuniform. You get a distribution of percentage points something like this:
image

It's impossible to express anything <18.18% or >81.82%, and the difference between ticks varies from 0.39% to 2.94%.  If the user doesn't understand this limitation, they could be off what they expect by 1.47% if they're in range and 18% if they're out of range.  Maybe 1.47% is small enough that it's fine and we just panic if they give us something out of range? That's deferring something that could be a compile-time check to runtime, though...

3) They can also get a significant difference in total clocks per cycle if off of a small duty cycle adjustment - here's the distribution of total clocks / duty cycle (assuming we use the lowest number of clocks that achieve a given duty cycle):
image

The worst case is the difference between a 48% duty cycle and a 49% duty cycle, which more than quadruples the number of clocks per cycle (4 -> 17), which I think cuts throughput by a factor of 4?  This sounds like it might be a foot-gun to me, but haven't done a ton of i2c stuff so if users are likely to be expecting this based on other implementations then it might be fine - what do you think?

jerrysxie (approved):

examples/rt685s-evk/src/bin/i2c-loopback-async-10bit.rs
williampMSFT: Update - talked to Felipe offline, I missed that we could also vary the clkdiv register to try to change the duration of a clock cycle, thereby accounting for the varying number of clocks that we spend high/low.  I'll look into doing that.

williampMSFT: Addressed this in the most recent iteration - we now compute the closest possible duty cycle and vary the clock divider to try to hit the requested (speed, duty cycle) tuple.

src/i2c/master.rs
```diff
@@ -31,6 +34,112 @@ pub enum Speed {
...
+    clock_div_multiplier: u16,
+
+    actual_duty_cycle: u8,
+    actual_freq_hz: u32,
+}
+
+impl SpeedRegisterSettings {
+    fn new(target_duty_cycle: u8, speed: Speed) -> Result<Self> {
+        const LO_VARIANTS: [Mstscllow; 8] = [
+            Mstscllow::Clocks2,
```
williampMSFT, line 53: is there a more idiomatic way to loop through enum variants like this? I couldn't find one but this seems a bit gnarly to me

williampMSFT: Yeah, I was hoping to do this without invoking unsafe code (and as far as I can tell, casting an integral type to an enum requires a core::mem::transmute, which is unsafe), but if that's the idiomatic way to do it then that's what I'll do.  Thanks!

williampMSFT: per our chat offline, can dodge unsafe code by duplicating the enum in a conversion function, which is about the same amount of code as this array nonsense I was trying, so did that

examples/rt685s-evk/src/bin/i2c-loopback-async.rs
```diff
@@ -108,7 +108,11 @@ async fn main(spawner: Spawner) {
 
     let slave = I2cSlave::new_async(p.FLEXCOMM2, p.PIO0_18, p.PIO0_17, Irqs, SLAVE_ADDR.unwrap(), p.DMA0_CH4).unwrap();
 
-    let master = I2cMaster::new_async(p.FLEXCOMM4, p.PIO0_29, p.PIO0_30, Irqs, Speed::Standard, p.DMA0_CH9).unwrap();
+    let config = i2c::master::Config {
+        speed: i2c::master::Speed::Standard,
+        duty_cycle: 40,
```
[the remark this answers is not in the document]
williampMSFT: Yeah, there are a couple other examples where we're just using the default for everything but I figured it might be valuable to have at least one example that demonstrates how to set them to something other than the default so I went with this one.

Per a conversation I had offline with Jerry, we want the default to remain the same as it was before, which was actually 40% duty cycle at Standard speed, so updated the default to that and changed this to be different from the default.

src/i2c/master.rs
```diff
@@ -31,6 +34,131 @@ pub enum Speed {
...
+
+        // Figure out what we need to set the clock divider to in order to hit the I2C speed the user requested.  Again, we may not
+        // be able to be exact, so we need to find the closest viable option.
+        //
+        let (result_clocks_hi, result_clocks_lo, result_div_multiplier) = (MIN_CLOCKS..=MAX_CLOCKS)
+            .cartesian_product(MIN_CLOCKS..=MAX_CLOCKS)
+            .filter(|(hi_clocks, lo_clocks)| get_duty_cycle(*hi_clocks, *lo_clocks) == closest_supported_duty_cycle)
+            .map(|(hi_clocks, lo_clocks)| {
+                let clock_div_multiplier =
+                    rounded_divide(SFRO_CLOCK_SPEED_HZ, target_freq_hz * u32::from(hi_clocks + lo_clocks)) as u16;
```
jerrysxie, line 135: So this would be rounding up to make sure that the divisor is never zero? It is a little hard to tell. But that seems to be case.

```
rounded_divide(16_000_000, 1_000_000 * (9+9))
= rounded_divide(16_000_000, 18_000_000)
= (16_000_000 + 18_000_000 / 2) / 18_000_000
= (16_000_000 + 9_000_000) / 18_000_000
= 25_000_000 / 18_000_000
= 1

```

src/i2c/master.rs
```diff
@@ -31,6 +34,131 @@ pub enum Speed {
...
+    scl_high_clocks: Mstsclhigh,
+    scl_low_clocks: Mstscllow,
+    clock_div_multiplier: u16,
+
+    _actual_duty_cycle: u8,
+    _actual_freq_hz: u32,
+}
+
+impl SpeedRegisterSettings {
+    fn new(target_duty_cycle: u8, speed: Speed) -> Result<Self> {
```
jerrysxie, line 100: It seems like the meat of this function can be extracted out to be unit-tested.

jerrysxie: Created an issue to do this: https://github.com/OpenDevicePartnership/embassy-imxrt/issues/407, non-blocking.

src/i2c/master.rs
```diff
@@ -39,13 +167,32 @@ pub struct I2cMaster<'a, M: Mode> {
...
 }
 
+/// Configuration for I2C Master
+#[derive(Clone, Copy)]
+pub struct Config {
+    /// Bus speed (nominal SCL, no clock stretching)
+    pub speed: Speed,
+
+    /// The target duty cycle (percentage of time to hold the SCL line high per bit).  Fitting is best-effort / not exact. Legal ranges are [18,82].
+    pub duty_cycle: u8,
```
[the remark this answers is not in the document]
williampMSFT: We could - I chatted with @felipebalbi on this last week and he suggested just rounding to the nearest possible duty cycle, which is what we're currently doing (so 0 will round to 18 and 100 will round to 82).  If you're in the (18, 82) range then the maximum error between requested duty cycle and actual duty cycle is 1.47%, but if you're outside that range it's 18%.

I'm happy to go either way here.  @felipebalbi any objections to disallowing construction of configs with out-of-range duty cycles?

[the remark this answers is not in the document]
williampMSFT: made a checked duty cycle type.  couldn't find an analogue to writing a constexpr function that throws to get compile-time checking for the validity of const duty cycles because I guess Result<T>::unwrap() isn't const? is there a more idiomatic way to do that, or do we just accept that the check has to be deferred to runtime for constant duty cycles?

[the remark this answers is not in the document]
RobertZ2011: The only way to do verification like this at compile time currently would be an external crate like `const_panic`. Neither functions like `unwrap` are const nor is `panic!` compatible with const code currently.

RobertZ2011 (approved):

felipebalbi (approved):

PR #399, opened 2025-04-16
Fix i2c remediation

jamesmunns:
In #367 I added remediation driven by the interrupt when a transaction was cancelled. However, I placed both the master/controller remediation and slave/target remediation under the `mstpending` event, where the slave remediation needs to be triggered when the `slvpending` event occurs.

This was caught by @davidv1992 on review.


davidv1992 (approved):
Seems correct to me.

jerrysxie (approved):

PR #401, opened 2025-04-17
FlexSPI code clean up

### LINKED ISSUE #364: FlexSPI issues found during review
https://github.com/OpenDevicePartnership/embassy-imxrt/pull/312

- [x] 1 - Use chunks to handle slices in read and write cmd APIs
- [x] 2 - Performance measurement to decide poll timeouts is the FlexSPI code
- [x] 3 - Fix the IO pin handling and remove Options to handle right ownership
- [x] 4 - Refactor configure_flexspi_device_port_a and configure_flexspi_device_port_b functions to extract common parts
- [x] 5 - Make RX and TX watermark user configurable
- [x] 6 - Replace new_blocking() parameters with a config structure
- [x] 7 - Handle Line breaks at 120 characters
- [x] 8 - Using CmdResult for Error handling struct (nit: How about defining another struct CmdResult which contains these values? Then you reduce some of the repetition.)
- [x] 9 - Refactor check_timeout() function as per suggestion - (this looks like a fairly generic helper, maybe add it to lib.rs with a pub(crate)?
- [x] I would rename this to something like expired()
- [x] You don't need the if:
- [x] fn expired(start: Instant, timeout: u64) -> bool {
- [x]     Instant::now().duration_since(start) > timeout
- [x] })
- [x] 10 - Deriving defmt for describe function
- [x] 11 - impl From<FlexSpiError> for NorStorageBusError
- [x] 12 - read_buf refactor (let buffer = read_buf.ok_or(NorStorageBusError::StorageBusInternalError)?;
- [x] self.read_data(cmd, buffer)?;)
- [x] 13 - Checkout the suggestions by Felipe in the PR where this issue is referred for taking up the suggestion later
- [x] 14 - Restructure flexspi files as per suggestion - _We should have two files: flexspi/mod.rs with all the necessary definitions and flexspi/nor.rs for the NOR flash driver. Something similar to this HAL for iMXRT1xxx devices: https://github.com/korbin/embassy-imxrt/blob/main/src/flexspi/nor.rs_
- [ ]

Shaibal-Microsoft:
This PR does the cleanup as per the issue

Closes #364

### DIFF
--- a/examples/rt685s-evk/src/bin/flexspi-storage-service.rs
+++ b/examples/rt685s-evk/src/bin/flexspi-storage-service.rs
@@ -3,9 +3,9 @@
 
 use defmt::info;
 use embassy_executor::Spawner;
-use embassy_imxrt::flexspi_nor_storage_bus::{
-    AhbConfig, FlexSpiBusWidth, FlexSpiFlashPort, FlexSpiFlashPortDeviceInstance, FlexspiAhbBufferConfig,
-    FlexspiConfig, FlexspiDeviceConfig, FlexspiNorStorageBus,
+use embassy_imxrt::flexspi::nor::{
+    AhbConfig, FlexSpiFlashPort, FlexSpiFlashPortDeviceInstance, FlexspiAhbBufferConfig, FlexspiConfig,
+    FlexspiConfigPortData, FlexspiDeviceConfig, FlexspiNorStorageBus,
 };
 use embassy_imxrt::pac::flexspi::ahbcr::*;
 use embassy_imxrt::pac::flexspi::flshcr1::*;
@@ -397,26 +397,27 @@ async fn main(_spawner: Spawner) {
         enable_same_config_for_all: Samedeviceen::Samedeviceen0,
         seq_timeout_cycle: 0xFFFF,
         ip_grant_timeout_cycle: 0xff,
-        tx_watermark: 0x08,
-        rx_watermark: 0x08,
         ahb_config,
     };
 
-    let mut flexspi_storage = FlexspiNorStorageBus::new_blocking(
-        p.FLEXSPI,       // FlexSPI peripheral
-        Some(p.PIO1_11), // FlexSPI DATA 0 pin
-        Some(p.PIO1_12),
-        Some(p.PIO1_13),
-        Some(p.PIO1_14),
-        Some(p.PIO2_17),
-        Some(p.PIO2_18),
-        Some(p.PIO2_22),
-        Some(p.PIO2_23),
+    let mut flexspi_storage = FlexspiNorStorageBus::new_blocking_octal_config(
+        p.FLEXSPI, // FlexSPI peripheral
+        p.PIO1_11,
+        p.PIO1_12,
+        p.PIO1_13,
+        p.PIO1_14,
+        p.PIO2_17,
+        p.PIO2_18,
+        p.PIO2_22,
+        p.PIO2_23,
         p.PIO1_29,
         p.PIO2_19,
-        FlexSpiFlashPort::PortB,                         // FlexSPI port
-        FlexSpiBusWidth::Octal,                          // FlexSPI bus width
-        FlexSpiFlashPortDeviceInstance::DeviceInstance0, // FlexSPI device instance
+        FlexspiConfigPortData {
+            port: FlexSpiFlashPort::PortB,                                 // FlexSPI port
+            dev_instance: FlexSpiFlashPortDeviceInstance::DeviceInstance0, // FlexSPI device instance
+            rx_watermark: 0x8,
+            tx_watermark: 0x8,
+        },
     );
 
     // Configure the Flexspi controller
--- a/src/flexspi/mod.rs
+++ b/src/flexspi/mod.rs
@@ -1,0 +1,9 @@
+pub mod nor;
+
+#[cfg(feature = "time")]
+use embassy_time::Instant;
+
+#[cfg(feature = "time")]
+pub(crate) fn is_expired(start: Instant, timeout: u64) -> bool {
+    Instant::now().duration_since(start).as_millis() > timeout
+}
--- a/src/flexspi/nor.rs
+++ b/src/flexspi/nor.rs
@@ -7,12 +7,15 @@ use embassy_hal_internal::{Peri, PeripheralType};
 use embassy_time::Instant;
 use mimxrt600_fcb::FlexSpiLutOpcode;
 use mimxrt600_fcb::FlexSpiLutOpcode::*;
+use paste::paste;
 use storage_bus::nor::{
     BlockingNorStorageBusDriver, NorStorageBusError, NorStorageBusWidth, NorStorageCmd, NorStorageCmdMode,
     NorStorageCmdType, NorStorageDummyCycles,
 };
 
 use crate::clocks::enable_and_reset;
+#[cfg(feature = "time")]
+use crate::flexspi::is_expired;
 use crate::iopctl::IopctlPin as Pin;
 use crate::pac::flexspi::ahbcr::*;
 use crate::pac::flexspi::flshcr1::*;
@@ -22,20 +25,87 @@ use crate::pac::flexspi::mcr0::*;
 use crate::pac::flexspi::mcr2::*;
 use crate::{interrupt, peripherals};
 
-const MAX_FLEXSPI_TRANSFER_SIZE: u32 = 128;
-const FLEXSPI_OP_SEQ_NUMBER: u8 = 0;
-const FLEXSPI_LUT_UNLOCK_CODE: u32 = 0x5AF05AF0;
+macro_rules! configure_ports_a {
+    ($port:expr, $regs: ident, $device_config: ident, $flash_size: ident) => {
+        paste! {
+            $regs.[<flsha $port cr0>]().modify(|_, w| unsafe { w.flshsz().bits($flash_size) });
+            $regs.[<flshcr1a $port>]().modify(|_, w| unsafe {
+                w.csinterval()
+                    .bits($device_config.cs_interval)
+                    .tcsh()
+                    .bits($device_config.cs_hold_time)
+                    .tcss()
+                    .bits($device_config.cs_setup_time)
+                    .cas()
+                    .bits($device_config.columnspace)
+                    .wa()
+                    .bit($device_config.enable_word_address)
+                    .csintervalunit()
+                    .variant($device_config.cs_interval_unit)
+            });
+            $regs.[<flshcr2a $port>]()
+                .modify(|_, w| w.awrwaitunit().variant($device_config.ahb_write_wait_unit));
+
+            if $device_config.ard_seq_number > 0 {
+                $regs.[<flshcr2a $port>]().modify(|_, w| unsafe {
+                    w.ardseqnum()
+                        .bits($device_config.ard_seq_number - 1)
+                        .ardseqid()
+                        .bits($device_config.ard_seq_index)
+                });
+            }
+        }
+    };
+}
+
+macro_rules! configure_ports_b {
+    ($port:expr, $regs: ident, $device_config: ident, $flash_size: ident) => {
+        paste! {
+            $regs.[<flshb $port cr0>]().modify(|_, w| unsafe { w.flshsz().bits($flash_size) });
+            $regs.[<flshcr1b $port>]().modify(|_, w| unsafe {
+                w.csinterval()
+                    .bits($device_config.cs_interval)
+                    .tcsh()
+                    .bits($device_config.cs_hold_time)
+                    .tcss()
+                    .bits($device_config.cs_setup_time)
+                    .cas()
+                    .bits($device_config.columnspace)
+                    .wa()
+                    .bit($device_config.enable_word_address)
+                    .csintervalunit()
+                    .variant($device_config.cs_interval_unit)
+            });
+            $regs.[<flshcr2b $port>]()
+                .modify(|_, w| w.awrwaitunit().variant($device_config.ahb_write_wait_unit));
+
+            if $device_config.ard_seq_number > 0 {
+                $regs.[<flshcr2b $port>]().modify(|_, w| unsafe {
+                    w.ardseqnum()
+                        .bits($device_config.ard_seq_number - 1)
+                        .ardseqid()
+                        .bits($device_config.ard_seq_index)
+                });
+            }
+        }
+    };
+}
+
+const FIFO_SLOT_SIZE: u32 = 4; // 4 bytes
+const MAX_TRANSFER_SIZE: u32 = 128;
+const OPERATION_SEQ_NUMBER: u8 = 0;
+const LUT_UNLOCK_CODE: u32 = 0x5AF05AF0;
 
 #[cfg(feature = "time")]
-const FLEXSPI_CMD_COMPLETION_TIMEOUT: u64 = 1000; // 1 second
+const CMD_COMPLETION_TIMEOUT: u64 = 10; // 10 millisecond
 #[cfg(feature = "time")]
-const FLEXSPI_DATA_FILL_TIMEOUT: u64 = 1000; // 1 second
+const DATA_FILL_TIMEOUT: u64 = 10; // 10 millisecond
 #[cfg(feature = "time")]
-const FLEXSPI_TX_FIFO_FREE_WATERMARK_TIMEOUT: u64 = 1000; // 1 second
+const TX_FIFO_FREE_WATERMARK_TIMEOUT: u64 = 10; // 10 millisecond
 #[cfg(feature = "time")]
-const FLEXSPI_RESET_TIMEOUT: u64 = 1000; // 1 second
+const RESET_TIMEOUT: u64 = 10; // 10 millisecond
 #[cfg(feature = "time")]
-const FLEXSPI_IDLE_TIMEOUT: u64 = 1000; // 1 second
+const IDLE_TIMEOUT: u64 = 10; // 10 millisecond
 
 const CLOCK_100MHZ: u32 = 100_000_000;
 const DELAYCELLUNIT: u32 = 75; // 75ps
@@ -56,6 +126,18 @@ pub enum FlexSpiFlashPortDeviceInstance {
     DeviceInstance0,
     /// Device Instance 1
     DeviceInstance1,
+}
+
+/// FlexSPI Configuration Port data structure
+pub struct FlexspiConfigPortData {
+    /// FlexSPI Port - PortA or PortB
+    pub port: FlexSpiFlashPort,
+    /// FlexSPI Flash Port Device Instance - DeviceInstance0 or DeviceInstance1
+    pub dev_instance: FlexSpiFlashPortDeviceInstance,
+    /// RX watermark level
+    pub rx_watermark: u8,
+    /// TX watermark level
+    pub tx_watermark: u8,
 }
 
 #[derive(Clone, Copy, Debug)]
@@ -122,7 +204,8 @@ pub struct FlexspiAhbBufferConfig {
     pub master_index: u8,
     /// AHB buffer size in byte.   
     pub buffer_size: u16,
-    /// AHB Read Prefetch Enable for current AHB RX Buffer corresponding Master, allows to prefetch data for AHB read access.
+    /// AHB Read Prefetch Enable for current AHB RX Buffer corresponding Master, allows to prefetch
+    /// data for AHB read access.
     pub enable_prefetch: bool,
 }
 
@@ -178,17 +261,20 @@ pub struct AhbConfig {
     pub ahb_grant_timeout_cycle: u8,
     /// Timeout wait cycle for AHB read/write access, timeout after ahbBusTimeoutCycle*1024 AHB clock cycles.
     pub ahb_bus_timeout_cycle: u16,
-    /// Wait cycle for idle state before suspended command sequence resume, timeout after ahbBusTimeoutCycle AHB clock cycles.
+    /// Wait cycle for idle state before suspended command sequence resume, timeout after ahbBusTimeoutCycle
+    /// AHB clock cycles.
     pub resume_wait_cycle: u8,
     /// AHB buffer size.
     pub buffer: [FlexspiAhbBufferConfig; 8],
     /// Enable/disable automatically clean AHB RX Buffer and TX Buffer when FLEXSPI returns STOP mode ACK.
     pub enable_clear_ahb_buffer_opt: Clrahbbufopt,
-    /// Enable/disable remove AHB read burst start address alignment limitation. when enable, there is no AHB read burst start address alignment limitation.
+    /// Enable/disable remove AHB read burst start address alignment limitation. when enable, there is no AHB
+    /// read burst start address alignment limitation.
     pub enable_read_address_opt: Readaddropt,
     /// Enable/disable AHB read prefetch feature, when enabled, FLEXSPI will fetch more data than current AHB burst.
     pub enable_ahb_prefetch: bool,
-    /// Enable/disable AHB bufferable write access support, when enabled, FLEXSPI return before waiting for command execution finished.
+    /// Enable/disable AHB bufferable write access support, when enabled, FLEXSPI return before waiting for command
+    /// execution finished.
     pub enable_ahb_bufferable: Bufferableen,
     /// Enable AHB bus cachable read access support.
     pub enable_ahb_cachable: Cachableen,
@@ -201,24 +287,24 @@ pub struct FlexspiConfig {
     pub rx_sample_clock: Rxclksrc,
     /// Enable/disable SCK output free-running.
     pub enable_sck_free_running: Sckfreerunen,
-    /// Enable/disable combining PORT A and B Data Pins (SIOA[3:0] and SIOB[3:0]) to support Flash Octal mode.
+    /// Enable/disable combining PORT A and B Data Pins (SIOA[3:0] and SIOB[3:0]) to support
+    /// Flash Octal mode.
     pub enable_combination: bool,
     /// Enable/disable doze mode support.
     pub enable_doze: Dozeen,
     /// Enable/disable divide by 2 of the clock for half speed commands.
     pub enable_half_speed_access: Hsen,
-    /// Enable/disable SCKB pad use as SCKA differential clock output, when enable, Port B flash access is not available.
+    /// Enable/disable SCKB pad use as SCKA differential clock output, when enable, Port B flash access
+    /// is not available.
     pub enable_sck_b_diff_opt: Sckbdiffopt,
-    /// Enable/disable same configuration for all connected devices when enabled, same configuration in FLASHA1CRx is applied to all.
+    /// Enable/disable same configuration for all connected devices when enabled, same configuration in
+    /// FLASHA1CRx is applied to all.
     pub enable_same_config_for_all: Samedeviceen,
-    /// Timeout wait cycle for command sequence execution, timeout after ahbGrantTimeoutCyle*1024 serial root clock cycles.
+    /// Timeout wait cycle for command sequence execution, timeout after ahbGrantTimeoutCyle*1024 serial
+    /// root clock cycles.
     pub seq_timeout_cycle: u16,
     /// Timeout wait cycle for IP command grant, timeout after ipGrantTimeoutCycle*1024 AHB clock cycles.
     pub ip_grant_timeout_cycle: u8,
-    /// FLEXSPI IP transmit watermark value.
-    pub tx_watermark: usize,
-    /// FLEXSPI receive watermark value.
-    pub rx_watermark: usize,
     /// AHB configuration
     pub ahb_config: AhbConfig,
 }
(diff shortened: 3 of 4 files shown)


src/flexspi/nor.rs
```diff
@@ -22,20 +25,86 @@ use crate::pac::flexspi::mcr0::*;
 use crate::pac::flexspi::mcr2::*;
 use crate::{interrupt, peripherals};
 
+macro_rules! configure_ports_a {
```
jerrysxie, line 28: Since this is a refactoring PR, it makes sense to break it up into multiple logical commits. Is there any reason why we squash all these commits into a single commit?

Shaibal-Microsoft: I did not create any commit history initially. There is no logical commits for me to refer. I planned to create separate commits during review.
I can setup a review call with you if you like

jerrysxie (changes requested):

src/flexspi/nor.rs
```diff
@@ -22,20 +25,87 @@ use crate::pac::flexspi::mcr0::*;
...
+}
+
+const FLEXSPI_FIFO_ENTRY_SIZE: u32 = 4; // 4 bytes
 const MAX_FLEXSPI_TRANSFER_SIZE: u32 = 128;
 const FLEXSPI_OP_SEQ_NUMBER: u8 = 0;
 const FLEXSPI_LUT_UNLOCK_CODE: u32 = 0x5AF05AF0;
 
 #[cfg(feature = "time")]
-const FLEXSPI_CMD_COMPLETION_TIMEOUT: u64 = 1000; // 1 second
+const FLEXSPI_CMD_COMPLETION_TIMEOUT: u64 = 10; // 10 millisecond
```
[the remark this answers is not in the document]
Shaibal-Microsoft: fixed

src/flexspi/nor.rs
```diff
@@ -538,66 +699,34 @@ impl<'d> BlockingNorStorageBusDriver for FlexspiNorStorageBus<'d, Blocking> {
...
-                FlexSpiError::DataLearningFailed => {
-                    return Err(NorStorageBusError::StorageBusInternalError);
-                }
-            }
-        }
+        self.check_transfer_status().map_err(|e| {
+            e.describe(self);
+            <FlexSpiError as Into<FlexSpiError>>::into(e)
+        })?;
+        // if let Err(status) = self.check_transfer_status() {
```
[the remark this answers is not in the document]
Shaibal-Microsoft: fixed

src/flexspi/nor.rs
```diff
@@ -945,33 +1081,30 @@ impl<'d, M: Mode> FlexspiNorStorageBus<'d, M> {
 
 impl<'d> FlexspiNorStorageBus<'d, Blocking> {
     fn read_data(&mut self, cmd: NorStorageCmd, read_buf: &mut [u8]) -> Result<(), NorStorageBusError> {
-        if let Some(size) = cmd.data_bytes {
-            if read_buf.len() != size as usize {
-                return Err(NorStorageBusError::StorageBusInternalError);
-            }
+        let size = cmd.data_bytes.ok_or(NorStorageBusError::StorageBusInternalError)?;
```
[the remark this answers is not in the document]
Shaibal-Microsoft: I agree. I think InvalidParameter makes more sense. Right now we dont have any other type better suited.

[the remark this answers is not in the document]
Shaibal-Microsoft: Filed https://github.com/OpenDevicePartnership/embedded-services/issues/258 for this

src/flexspi/nor.rs
```diff
@@ -1117,47 +1251,14 @@ impl<'d> FlexspiNorStorageBus<'d, Blocking> {
...
-                self.info
-                    .regs
-                    .tfdr(num_fifo_slot as usize)
-                    .write(|w| unsafe { w.bits(temp) });
+            for (chunk, slot) in chunk_iter
+                .chunks(FLEXSPI_FIFO_ENTRY_SIZE as usize)
+                .zip(0..num_tx_watermark_slot)
+            {
+                let temp: u32 = 0;
+                temp.to_le_bytes().copy_from_slice(chunk);
```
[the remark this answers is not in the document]
Shaibal-Microsoft: write operation is functional. I am not sure then how its working because temp is 0 as per what you mentioned. Its weird !!

src/flexspi/nor.rs
```diff
@@ -1433,250 +1515,108 @@ impl FlexSpiConfigurationPort {
...
+            FlexspiPinConfig::QuadSpiPin { data_pins, clk, cs } => {
+                data_pins.data0.config_pin();
+                data_pins.data1.config_pin();
+                data_pins.data2.config_pin();
+                data_pins.data3.config_pin();
+                clk.config_pin();
+                cs.config_pin();
+            }
+            FlexspiPinConfig::OctalSpiPin { data_pins, clk, cs } => {
+                data_pins.data0.config_pin();
```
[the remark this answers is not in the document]
Shaibal-Microsoft: Not sure what you meant by manual. I did not see any section related to pin configuration in the manual. I chose the configuration based on previous program where HW team shared the configuration.

Glitches will come when we are doing XIP. If we are not doing that, then we dont need to worry about glitches.

Not all pins will be configured by BootROM. BoorROM decides based on FCB which could have technically different bus width support than what user wants later post boot.

To give the user option for XIP boot, I added another config option called "RomPinConfig". When user passes this, we will ignore all pin configuration

src/flexspi/mod.rs
```diff
@@ -0,0 +1,9 @@
+pub mod nor;
+
+#[cfg(feature = "time")]
+use embassy_time::Instant;
+
+#[cfg(feature = "time")]
+pub(crate) fn expired(start: Instant, timeout: u64) -> bool {
```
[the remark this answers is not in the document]
Shaibal-Microsoft: fixed

src/flexspi/nor.rs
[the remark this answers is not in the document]
Shaibal-Microsoft: fixed. Did following - 

 let temp = u32::from_ne_bytes(
                    chunk
                        .try_into()
                        .map_err(|_| NorStorageBusError::StorageBusInternalError)?,
                );
                self.info.regs.tfdr(slot as usize).write(|w| unsafe { w.bits(temp) });

src/flexspi/nor.rs
```diff
@@ -1096,17 +1224,23 @@ impl<'d> FlexspiNorStorageBus<'d, Blocking> {
...
+        let mut slot_group = num_fifo_slot / num_tx_watermark_slot as u32;
 
-        for _ in 0..slot_group {
+        if num_fifo_slot % num_tx_watermark_slot as u32 != 0 {
+            // Data length is not multiple of tx_watermark. We want to have 1 more slot to write the data
+            // in the last slot. Hence increment the slot group by 1
+            slot_group += 1;
+        }
+
+        for (chunk_iter, _slot) in write_data.chunks(self.tx_watermark as usize).zip(0..slot_group) {
```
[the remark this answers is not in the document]
Shaibal-Microsoft: fixed

src/flexspi/nor.rs
```diff
@@ -1006,17 +1139,17 @@ impl<'d> FlexspiNorStorageBus<'d, Blocking> {
             e.describe(self);
             return Err(NorStorageBusError::StorageBusIoError);
         }
-        num_fifo_slot = size / 4;
-        num_rx_watermark_slot = self.rx_watermark / 4;
+
+        num_fifo_slot = size / FLEXSPI_FIFO_ENTRY_SIZE as u32;
```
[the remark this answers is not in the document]
Shaibal-Microsoft: fixed

src/flexspi/mod.rs
```diff
@@ -0,0 +1,9 @@
...
+    Instant::now().duration_since(start).as_millis() > timeout
```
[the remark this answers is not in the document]
Shaibal-Microsoft: @jerrysxie  Implicit casting should take care of this right?

src/flexspi/nor.rs
```diff
@@ -1027,86 +1158,78 @@ impl<'d> FlexspiNorStorageBus<'d, Blocking> {
...
         self.info.regs.intr().modify(|_, w| w.iprxwa().clear_bit_by_one());
 
         Ok(())
     }
 
-    fn write_cmd_data(&mut self, mut size: u32, write_data: &[u8]) -> Result<(), NorStorageBusError> {
-        let mut num_fifo_slot;
-        let mut byte_cnt = 0;
-
+    fn write_cmd_data(&mut self, _size: u32, write_data: &[u8]) -> Result<(), NorStorageBusError> {
```
jerrysxie, line 1216: It seems like `size` is not even needed for this function if we are relying on `write_data.len()` only, and this is not a public interface, right?

src/flexspi/nor.rs
```diff
@@ -996,27 +1122,32 @@ impl<'d> FlexspiNorStorageBus<'d, Blocking> {
 
     fn read_cmd_data(&mut self, mut size: u32, read_data: &mut [u8]) -> Result<(), NorStorageBusError> {
```
jerrysxie, line 1123: `size` does not seem like it is needed for this function as well in its usage.

jerrysxie (commented):
What about item 10 in #364? What does it refer to?

Shaibal-Microsoft:
> What about item 10 in #364? What does it refer to?

I missed this I guess. Following comment from Felipe was for item 10.

image


I didn't understand what was the suggestion. Could you please help? @felipebalbi

src/flexspi/nor.rs
Shaibal-Microsoft: agree.

src/flexspi/nor.rs
Shaibal-Microsoft: agree

felipebalbi:
> > What about item 10 in #364? What does it refer to?
> 
> 
> 
> I missed this I guess. Following comment from Felipe was for item 10.
> 
> 
> 
> image
> 
> 
> 
> 
> 
> I didn't understand what was the suggestion. Could you please help? @felipebalbi

Something like this:

```rust
#[cfg_attr(feature = "defmt", derive(defmt::Format)]
```

Shaibal-Microsoft:
> > > What about item 10 in #364? What does it refer to?
> > 
> > 
> > I missed this I guess. Following comment from Felipe was for item 10.
> > image
> > I didn't understand what was the suggestion. Could you please help? @felipebalbi
> 
> Something like this:
> 
> ```rust
> #[cfg_attr(feature = "defmt", derive(defmt::Format)]
> ```

Looks like I already did that. Is this what you wanted @felipebalbi ?

image

examples/rt685s-evk/src/bin/flexspi-storage-service.rs
```diff
@@ -3,9 +3,9 @@
 
 use defmt::info;
 use embassy_executor::Spawner;
-use embassy_imxrt::flexspi_nor_storage_bus::{
+use embassy_imxrt::flexspi::nor::{
```
Shaibal-Microsoft, line 6: @felipebalbi @jerrysxie Since we are pushing the code to ODP, we could have user with NAND or Hyperbus storage.
Do you think it makes sense to create issues for adding support for NAND and Hyperbus as a separate module with feature flags guarded? But the only problem is validation.

jerrysxie: Until we have an OEM come to us with an actual need for NAND and Hyperbus with a test platform. We can punt on these for now.

jerrysxie (dismissed):

src/flexspi/mod.rs
```diff
@@ -0,0 +1,9 @@
+pub mod nor;
+
+#[cfg(feature = "time")]
+use embassy_time::Instant;
+
+#[cfg(feature = "time")]
+pub(crate) fn is_expired(start: Instant, timeout: u64) -> bool {
+    Instant::now().duration_since(start).as_millis() > timeout
```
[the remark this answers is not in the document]
Shaibal-Microsoft: not required.

src/flexspi/nor.rs
```diff
@@ -994,29 +1120,35 @@ impl<'d> FlexspiNorStorageBus<'d, Blocking> {
...
 
-        for _ in 0..slot_group {
-            // Wait for RX FIFO to be filled with water mark level data
+        total_fifo_slot = size / FIFO_SLOT_SIZE as u32;
+        num_rx_watermark_slot = self.rx_watermark / FIFO_SLOT_SIZE as u8;
+        num_rx_watermark_chunks = total_fifo_slot / num_rx_watermark_slot as u32;
+
+        for (watermark_sized_chunk, _) in read_data
+            .chunks_mut(self.rx_watermark as usize)
+            .zip(0..num_rx_watermark_chunks)
```
JamesHuard, line 1143: why do you zip this if we're just throwing away the other element?

src/flexspi/nor.rs
```diff
@@ -607,18 +729,10 @@ impl<'d> BlockingNorStorageBusDriver for FlexspiNorStorageBus<'d, Blocking> {
...
-
-            None => {
-                // SAFETY: Operation is safe as we are programming 0 as default the address
-                self.info.regs.ipcr0().modify(|_, w| unsafe { w.sfar().bits(0) });
-            }
-        }
+        self.info
+            .regs
+            .ipcr0()
+            .modify(|_, w| unsafe { w.sfar().bits(addr.unwrap_or(0)) });
```
JamesHuard, line 735: missing // SAFETY: comment

src/flexspi/nor.rs
```diff
@@ -1027,86 +1159,78 @@ impl<'d> FlexspiNorStorageBus<'d, Blocking> {
...
-            while (self.info.regs.iprxfsts().read().fill().bits() * 8) < size as u8 {}
-        }
 
+        // Data was not rx_watermark aligned. We need to read the remaining data in the RX FIFO
         if size > 0 {
+            // Wait for the remaining bytes in RX FIFO to be filled which is less than rx_watermark
+            #[cfg(feature = "time")]
+            {
+                let start = Instant::now();
+                while (self.info.regs.iprxfsts().read().fill().bits() * 8) < size as u8 {
```
JamesHuard, line 1181: if size is a multiple of 256 won't this loop forever?

src/flexspi/nor.rs
```diff
@@ -1027,86 +1159,78 @@ impl<'d> FlexspiNorStorageBus<'d, Blocking> {
...
+                    let timedout = is_expired(start, DATA_FILL_TIMEOUT);
+                    if timedout {
+                        return Err(NorStorageBusError::StorageBusInternalError);
+                    }
+                }
+            }
+            #[cfg(not(feature = "time"))]
+            {
+                while (self.info.regs.iprxfsts().read().fill().bits() * 8) < size as u8 {}
```
JamesHuard, line 1190: same comment here

src/flexspi/nor.rs
JamesHuard: basically I think you need to reverse the cast or you're limiting size to < 256

JamesHuard (changes requested):
a few small questions, main concern is the two while () < size as u8

src/flexspi/nor.rs
Shaibal-Microsoft: We enter this block only when the **size is < watermark size** which cannot be more than 128 bytes as that is the total FIFO size.
watermark <= FIFO size

Hence down casting is not an issue here.

src/flexspi/nor.rs
Shaibal-Microsoft: same justification as above

src/flexspi/nor.rs
Shaibal-Microsoft: agree. we dont need zip here. will remove it

src/flexspi/nor.rs
Shaibal-Microsoft: fixed

src/flexspi/nor.rs
```diff
@@ -1117,47 +1241,17 @@ impl<'d> FlexspiNorStorageBus<'d, Blocking> {
...
+            for (chunk, slot) in watermark_sized_chunk
+                .chunks(FIFO_SLOT_SIZE as usize)
+                .zip(0..num_tx_watermark_slot)
+            {
+                let temp = u32::from_ne_bytes(
+                    chunk
+                        .try_into()
+                        .map_err(|_| NorStorageBusError::StorageBusInternalError)?,
+                );
+                self.info.regs.tfdr(slot as usize).write(|w| unsafe { w.bits(temp) });
```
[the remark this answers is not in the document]
Shaibal-Microsoft: fixed

JamesHuard (dismissed):

src/flexspi/nor.rs
Shaibal-Microsoft: so you are suggesting the following?

while (self.info.regs.iprxfsts().read().fill().bits() * 8) as u32 < size {

src/flexspi/nor.rs
jerrysxie: The read buffer is not guaranteed to be multiple of watermark-sized chunks. Previously `zip()` was short-circuiting the loop to only deal with the watermark-sized chunks and leave the last remainder chunk to be handle by the code afterwards. If we are going to remove the `zip()`, then we have to move the remainder chunk handling logic into this loop as well as a case when `chunk.len() < self.rx_watermark`

I asked Shaibal to add a comment to indicate this, because I had to stare at this for a long time to figure out the logic here:
```
            // If the size is not multiple of rx_watermark, we will break out of the loop without reading the remaining data
            // We need to read the remaining data in the RX FIFO which is less than rx_watermark
```

But it still seems to be hard to read, so maybe we should move away from `zip()` but move this logic into the loop when `chunk.len() < self.rx_watermark`:

```
        // Data was not rx_watermark aligned. We need to read the remaining data in the RX FIFO
        if size > 0 {
            // Wait for the remaining bytes in RX FIFO to be filled which is less than rx_watermark
            #[cfg(feature = "time")]
            {
                let start = Instant::now();
                while (self.info.regs.iprxfsts().read().fill().bits() * 8) < size as u8 {
                    let timedout = is_expired(start, DATA_FILL_TIMEOUT);
                    if timedout {
                        return Err(NorStorageBusError::StorageBusInternalError);
                    }
                }
            }
            #[cfg(not(feature = "time"))]
            {
                while (self.info.regs.iprxfsts().read().fill().bits() * 8) < size as u8 {}
            }
            // size must be between 1 and rx_watermark by now
            let mut temp;
            total_fifo_slot = size / FIFO_SLOT_SIZE;

            if size % FIFO_SLOT_SIZE != 0 {
                // Data length is not multiple of FLEXSPI_FIFO_ENTRY_SIZE. We want to have 1 more slot to read the data
                // in the last slot. Hence increment the slot group by 1
                total_fifo_slot += 1;
            }

            for (chunk, slot) in read_data[bytes_read as usize..]
                .chunks_mut(FIFO_SLOT_SIZE as usize)
                .zip(0..total_fifo_slot)
            {
                temp = self.info.regs.rfdr(slot as usize).read().bits();
                chunk.copy_from_slice(&temp.to_le_bytes());
            }
        }
```

src/flexspi/nor.rs
JamesHuard: exactly -- this will never lead to the infinite loop as long as size > 0, even if someone tweaks the logic later on

src/flexspi/nor.rs
Shaibal-Microsoft: agree. I can see the readability issue here.

JamesHuard: Thanks for catching this @jerrysxie. That detail completely eluded me

src/flexspi/nor.rs
Shaibal-Microsoft: fixed

src/flexspi/nor.rs
Shaibal-Microsoft: fixed

JamesHuard (dismissed):

src/flexspi/nor.rs
```diff
@@ -1136,75 +1134,49 @@ impl<'d> FlexspiNorStorageBus<'d, Blocking> {
...
-            #[cfg(not(feature = "time"))]
-            {
-                while (self.info.regs.iprxfsts().read().fill().bits() * 8) < size as u8 {}
-            }
-            // size must be between 1 and rx_watermark by now
-            let mut temp;
-            total_fifo_slot = size / FIFO_SLOT_SIZE;
-
-            if size % FIFO_SLOT_SIZE != 0 {
-                // Data length is not multiple of FLEXSPI_FIFO_ENTRY_SIZE. We want to have 1 more slot to read the data
```
jerrysxie: We removed the logic to deal with remainder bytes that is not a multiple of FIFO_SLOT_SIZE. But now that that I look at the code more closely. I realized that even the existing code will not work correctly, because in the case where `chunk.len() < FIFO_SLOT_SIZE`, `copy_from_slice()` will panic: https://doc.rust-lang.org/src/core/slice/mod.rs.html#3730. We need to address this somehow. If we are not hitting this panic right now, that means we are not testing with buffer size that is not a multiple of `FIFO_SLOT_SIZE` and not a multiple of `rx_watermark`.

We can add check to ensure that buffer size is a multiple of `FIFO_SLOT_SIZE` or we would have to add code to perform an extra fifo read but only copy 1-3 bytes from the u32 read value.

jerrysxie (changes requested):

src/flexspi/nor.rs
```diff
@@ -994,119 +1120,89 @@ impl<'d> FlexspiNorStorageBus<'d, Blocking> {
...
-            if size > 0 {
-                // size must be less than 4 bytes by now
-                temp = self.info.regs.rfdr(num_fifo_slot as usize).read().bits();
-                for j in 0..size {
-                    read_data[bytes_read as usize] = (temp >> (8 * j)) as u8;
-                    bytes_read += 1;
-                    size -= 1;
+            for (chunk, slot) in watermark_sized_chunk
+                .chunks_mut(FIFO_SLOT_SIZE as usize)
+                .zip(0..num_rx_watermark_slot)
```
jerrysxie, line 1170: We need to remove `zip()` call here, so we don't short-circuit and exit before handling the remainder chunk.

And if we remove the zip() call here. Some of the variables for tracking like `size` and  `num_rx_watermark_slot` might not be necessary anymore.

jerrysxie (changes requested):

src/flexspi/nor.rs
jerrysxie: Nevermind, did not realize that `num_rx_watermark_slot` is constant and did not scale to each `watermark_sized_chunk`, so it would short-cicuit when chunk is none.

jerrysxie (approved):
Pending testing with the latest round of changes, looks good to me.

Shaibal-Microsoft:
@JamesHuard  @jerrysxie 

I have changed the Flexspi object creation API. Now the API name contains the interface width and its much simpler.

Using 1 API to support all types was making it way too much complicated for supporting all the Generic pin types. Please note that I needed to support 8 pin types but then things will become complicated from user point of view when we dont want to use Octal interface.

I believe having interface width based API makes more sense and makes it super easy for users.

jerrysxie:
> @JamesHuard @jerrysxie
> 
> I have changed the Flexspi object creation API. Now the API name contains the interface width and its much simpler.
> 
> Using 1 API to support all types was making it way too much complicated for supporting all the Generic pin types. Please note that I needed to support 8 pin types but then things will become complicated from user point of view when we dont want to use Octal interface.
> 
> I believe having interface width based API makes more sense and makes it super easy for users.

What was the complication with the previous design? It looks perfectly fine to me. This new design makes it less ergonomic with long list of parameters. We should encapsulate the pins with a struct even if we want different `new()` for different configurations.

```
pub enum FlexspiPinConfig<'d, T: FlexSpiPin, CS: FlexSpiPin, CLK: FlexSpiPin> {
    /// Single SPI Pin Configuration
    SingleSpiPin {
        /// Single config
        data_pins: SpiSinglePin<'d, T>,
        /// SCK Pin
        clk: Peri<'d, CLK>,
        /// Chip select pin
        cs: Peri<'d, CS>,
    },
    /// Dual SPI Pin Configuration
    DualSpiPin {
        /// Dual Spi config
        data_pins: SpiDualPin<'d, T>,
        /// SCK Pin
        clk: Peri<'d, T>,
        /// Chip select pin
        cs: Peri<'d, T>,
    },
    /// Quad SPI Pin Configuration
    QuadSpiPin {
        /// Quad Spi config
        data_pins: SpiQuadPin<'d, T>,
        /// SCK Pin
        clk: Peri<'d, T>,
        /// Chip select pin
        cs: Peri<'d, T>,
    },
    /// Octal SPI Pin Configuration
    OctalSpiPin {
        /// Octal Spi config
        data_pins: SpiOctalPin<'d, T>,
        /// SCK Pin
        clk: Peri<'d, T>,
        /// Chip select pin
        cs: Peri<'d, T>,
    },
    /// XIP boot. We don't configure the pins as its configured by ROM code.
    RomPinConfig,
}
```

examples/rt685s-evk/src/bin/flexspi-storage-service.rs
```diff
@@ -397,26 +397,28 @@ async fn main(_spawner: Spawner) {
...
+        p.PIO2_22,
+        p.PIO2_23,
         p.PIO1_29,
         p.PIO2_19,
-        FlexSpiFlashPort::PortB,                         // FlexSPI port
-        FlexSpiBusWidth::Octal,                          // FlexSPI bus width
-        FlexSpiFlashPortDeviceInstance::DeviceInstance0, // FlexSPI device instance
+        FlexspiConfigPortData {
+            port: FlexSpiFlashPort::PortB,                                 // FlexSPI port
+            bus_width: FlexSpiBusWidth::Octal,                             // FlexSPI bus width
```
JamesHuard, line 417: is this redundant now with "octal config"?

src/flexspi/nor.rs
```diff
@@ -92,7 +92,7 @@ macro_rules! configure_ports_b {
 }
 
 const FIFO_SLOT_SIZE: u32 = 4; // 4 bytes
-const MAX_TRANSFER_SIZE: u32 = 128;
```
jerrysxie: Can we separate this into a separate PR? Let's not overload this PR with more and more changes.

jerrysxie (changes requested):

src/flexspi/nor.rs
Shaibal-Microsoft: yes I will do that

Shaibal-Microsoft:
> > @JamesHuard @jerrysxie
> > I have changed the Flexspi object creation API. Now the API name contains the interface width and its much simpler.
> > Using 1 API to support all types was making it way too much complicated for supporting all the Generic pin types. Please note that I needed to support 8 pin types but then things will become complicated from user point of view when we dont want to use Octal interface.
> > I believe having interface width based API makes more sense and makes it super easy for users.
> 
> What was the complication with the previous design? It looks perfectly fine to me. This new design makes it less ergonomic with long list of parameters. We should encapsulate the pins with a struct even if we want different `new()` for different configurations.
> 
> ```
> pub enum FlexspiPinConfig<'d, T: FlexSpiPin, CS: FlexSpiPin, CLK: FlexSpiPin> {
>     /// Single SPI Pin Configuration
>     SingleSpiPin {
>         /// Single config
>         data_pins: SpiSinglePin<'d, T>,
>         /// SCK Pin
>         clk: Peri<'d, CLK>,
>         /// Chip select pin
>         cs: Peri<'d, CS>,
>     },
>     /// Dual SPI Pin Configuration
>     DualSpiPin {
>         /// Dual Spi config
>         data_pins: SpiDualPin<'d, T>,
>         /// SCK Pin
>         clk: Peri<'d, T>,
>         /// Chip select pin
>         cs: Peri<'d, T>,
>     },
>     /// Quad SPI Pin Configuration
>     QuadSpiPin {
>         /// Quad Spi config
>         data_pins: SpiQuadPin<'d, T>,
>         /// SCK Pin
>         clk: Peri<'d, T>,
>         /// Chip select pin
>         cs: Peri<'d, T>,
>     },
>     /// Octal SPI Pin Configuration
>     OctalSpiPin {
>         /// Octal Spi config
>         data_pins: SpiOctalPin<'d, T>,
>         /// SCK Pin
>         clk: Peri<'d, T>,
>         /// Chip select pin
>         cs: Peri<'d, T>,
>     },
>     /// XIP boot. We don't configure the pins as its configured by ROM code.
>     RomPinConfig,
> }
> ```

@jerrysxie  As per discussion, going with this approach. The approach was originally suggested by @felipebalbi  in the first Flexspi PR. Thanks @felipebalbi

examples/rt685s-evk/src/bin/flexspi-storage-service.rs
Shaibal-Microsoft: yes. I will remove it

Shaibal-Microsoft: removed the bug fix

src/flexspi/nor.rs
Shaibal-Microsoft: fixed

jerrysxie (approved):

PR #412, opened 2025-05-01
Set MBINTALL by default for eSPI and allow sci to be cleared or set

philgweber:
The following is sequence of events for handling SCI

- MCU wants to assert SCI so we clear the SCIN bit in WIREWO
- I see VWire event 0x6 data 0x16 on the host indicating SCI event
- Host does a Peripheral Read to get the SCI related data
- Host sends interrupt to detail with the SCI internally on apps side
- Host sends a peripheral write to clear the SCI
- MCU receives peripheral write and writes SCIN --> 1 in WIREWO
- Host receives VWire event 0x6 data 0x17 indicating SCI is cleared
- WIREWO will read indicating SCIN  --> 1 only after host ACK's the VWire data

The MBINTALL must be set by default otherwise only the first read/write packets will generate interrupts.

### DIFF
--- a/src/espi.rs
+++ b/src/espi.rs
@@ -843,8 +843,8 @@ impl<'d> Espi<'d> {
     /// Active Low.
     ///
     /// Warning: Blocks until DONE bit clears
-    pub fn wake(&mut self) {
-        self.info.regs.wirewo().write(|w| w.waken_scin().clear_bit());
+    pub fn wake(&mut self, set: bool) {
+        self.info.regs.wirewo().write(|w| w.waken_scin().variant(!set));
         self.block_for_vwire_done();
     }
 
@@ -853,8 +853,8 @@ impl<'d> Espi<'d> {
     /// Active Low.
     ///
     /// Warning: Blocks until DONE bit clears
-    pub fn pme(&mut self) {
-        self.info.regs.wirewo().write(|w| w.pmen().clear_bit());
+    pub fn pme(&mut self, set: bool) {
+        self.info.regs.wirewo().write(|w| w.pmen().variant(!set));
         self.block_for_vwire_done();
     }
 
@@ -864,8 +864,8 @@ impl<'d> Espi<'d> {
     /// Active Low.
     ///
     /// Warning: Blocks until DONE bit clears
-    pub fn sci(&mut self) {
-        self.info.regs.wirewo().write(|w| w.scin().clear_bit());
+    pub fn sci(&mut self, set: bool) {
+        self.info.regs.wirewo().write(|w| w.scin().variant(!set));
         self.block_for_vwire_done();
     }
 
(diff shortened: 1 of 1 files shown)


src/espi.rs
```diff
@@ -864,8 +864,8 @@ impl<'d> Espi<'d> {
     /// Active Low.
     ///
     /// Warning: Blocks until DONE bit clears
-    pub fn sci(&mut self) {
-        self.info.regs.wirewo().write(|w| w.scin().clear_bit());
+    pub fn sci(&mut self, set: bool) {
```
jerrysxie, line 867: non-blocking, just for information: Is this behavior the same for other interrupt bit in `wirewo`?

src/espi.rs
```diff
@@ -1037,6 +1037,9 @@ impl Espi<'_> {
             .cfg()
             .modify(|_, w| w.direction().variant(direction));
 
+        // Enable MBINTAll by default
+        self.info.regs.port(port).cfg().modify(|_, w| w.mbint_all().set_bit());
```
jerrysxie, line 1041: We are doing 3 separate read-modify-write to port cfg register here. Can they be combined?

jerrysxie (changes requested):

src/espi.rs
philgweber: I was hoping the compiler would be smart enough to optimized that, but maybe it isn't with HW registers. If so I can combine this into a single op...

src/espi.rs
philgweber: Not all, but several others, I can change the others to read/wrote while I'm here :)

philgweber: Updated to include other bits in this register that need manual set and clear.

src/espi.rs
philgweber: Combined into a single write, also we really don't want to do modify here as it could be set to some other random values, reset all values when configuring the port.

src/espi.rs
```diff
@@ -843,8 +843,8 @@ impl<'d> Espi<'d> {
     /// Active Low.
     ///
     /// Warning: Blocks until DONE bit clears
-    pub fn wake(&mut self) {
-        self.info.regs.wirewo().write(|w| w.waken_scin().clear_bit());
+    pub fn wake(&mut self, set: bool) {
```
tullom, line 846: It may be worth writing a more descriptive note in the rustdocs for this and the subsequent active low functions that the parameter `set` is active low (but you pass in an active high value and the fn automatically inverts the value). Seems like a potential gotcha.

tullom (changes requested):

src/espi.rs
philgweber: Added further details in function headers to provide details on set and when we need to set true/false.

jerrysxie (approved):

felipebalbi (approved):

RobertZ2011 (approved):

tullom (approved):

PR #420, opened 2025-05-07
Pull audits from rust-crate-audits

akshat2112:
ODP rust crate audits are expected to be present in the rust-crate-audits repository. This PR enables importing those audits via `cargo vet`. This allows us to maintain a central audit repository, while using the same crates across ODP repositories.
The README is also being updated to reflect the change.


akshat2112:
Failing workflow can be ignored. It's failing due to race condition in the workflow that's fixed as part of the move to squash commits

felipebalbi (approved):

PR #421, opened 2025-05-07
Add support for using the OS Timer0 as time source

felipebalbi:
This PR lets the user choose between RTC and OS Timer0 as a time source.


felipebalbi:
There are some left-over `error!()` messages in `clocks.rs`. I'll fix those too.

felipebalbi:
Failures are false-positive. Workflow from latest commit is used for all commits.

src/time_driver.rs
```diff
@@ -1,4 +1,6 @@
+//! Time Driver.
```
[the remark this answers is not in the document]
felipebalbi: done

RobertZ2011 (dismissed):

felipebalbi:
Tagged with "Breaking Change" because once this is merged, the old `time-driver` feature is no more. User must select either `time-driver-rtc` or `time-driver-os-timer`. If need be, we can define `time-driver` to just enable one of those two and keep backwards compatibility.

src/time_driver/ostimer.rs
```diff
@@ -0,0 +1,155 @@
...
+        let t = self.now();
+        if timestamp <= t {
+            os().osevent_ctrl().modify(|_, w| w.ostimer_intena().clear_bit());
+            alarm.timestamp.set(u64::MAX);
+            return false;
+        }
+
+        let gray_timestamp = dec_to_gray(timestamp);
+
+        os().match_l()
```
jerrysxie, line 96: alt text

It seems like we have to wait for minimum of 3 bus clocks between updating the match registers. We can test the status register to see whether the write is safe. However, looking at the code, we do satisfy the minimum of 3 bus clock cycles between writes to match. Should we test the status bit to be safe?

jerrysxie (dismissed):

jerrysxie:
> Tagged with "Breaking Change" because once this is merged, the old `time-driver` feature is no more. User must select either `time-driver-rtc` or `time-driver-os-timer`. If need be, we can define `time-driver` to just enable one of those two and keep backwards compatibility.

I think we should define `time-driver` to just enable one of them as default to keep backwards compatibility.

src/time_driver/ostimer.rs
felipebalbi: There isn't much we can actually do. If we test, we add more latency which would result in a larger slack requirement. If the user passes a far too short timeout, it will just result in a "timeout in the past" but nothing catastrophic will happen. I think we keep it as is.

felipebalbi:
> > Tagged with "Breaking Change" because once this is merged, the old `time-driver` feature is no more. User must select either `time-driver-rtc` or `time-driver-os-timer`. If need be, we can define `time-driver` to just enable one of those two and keep backwards compatibility.
> 
> We should define `time-driver` to just enable one of them as default to keep backwards compatibility.

Defaulted to RTC as it was before. No impact to users.

jerrysxie (approved):

PR #422, opened 2025-05-08
Fix bug for packet size > 88 bytes

Shaibal-Microsoft:


src/flexspi/nor.rs
```diff
@@ -987,9 +1014,7 @@ impl<'d> FlexspiNorStorageBus<'d, Blocking> {
             return Err(NorStorageBusError::StorageBusInternalError);
         }
 
-        for chunk in read_buf.chunks_mut(MAX_TRANSFER_SIZE as usize) {
-            self.read_cmd_data(chunk)?;
```
Shaibal-Microsoft, line 991: We dont need to send in chunk as we can we are already checking the size of the packet before calling this function. Our HW FIFO supports MAX_READ_SIZE which we can send all together

src/flexspi/nor.rs
```diff
@@ -617,19 +618,45 @@ impl<'d> BlockingNorStorageBusDriver for FlexspiNorStorageBus<'d, Blocking> {
             <FlexSpiError as Into<FlexSpiError>>::into(e)
         })?;
 
-        // For data transfer commands, read/write the data
         if let Some(data_cmd) = cmd.cmdtype {
             match data_cmd {
                 NorStorageCmdType::Read => {
                     let buffer = read_buf.ok_or(NorStorageBusError::StorageBusInternalError)?;
+                    if cmd.data_bytes.unwrap() > MAX_RECEIVE_SIZE {
+                        return Err(NorStorageBusError::StorageBusInternalError);
```
felipebalbi, line 626: How about creating a dedicated error here?

```rust
return Err(NorStorageBusError::TransferTooLarge);
```

src/flexspi/nor.rs
```diff
@@ -617,19 +618,45 @@ impl<'d> BlockingNorStorageBusDriver for FlexspiNorStorageBus<'d, Blocking> {
...
+                    }
                     self.read_data(cmd, buffer)?;
                 }
                 NorStorageCmdType::Write => {
                     let buffer = write_buf.ok_or(NorStorageBusError::StorageBusInternalError)?;
+                    if cmd.data_bytes.unwrap() > MAX_TRANSMIT_SIZE {
+                        return Err(NorStorageBusError::StorageBusInternalError);
```
felipebalbi, line 633: ditto

src/flexspi/nor.rs
```diff
@@ -677,7 +704,7 @@ impl<'d, M: Mode> FlexspiNorStorageBus<'d, M> {
         if let Some(size) = size {
             self.info.regs.ipcr1().modify(|_, w| unsafe {
                 // SAFETY: Operation is safe as we are programming the size of the transfer
-                w.idatsz().bits(min(size, MAX_TRANSFER_SIZE) as u16)
+                w.idatsz().bits(size as u16)
```
felipebalbi, line 707: There's a potential for truncation here. Is that really what you wanted? What if user passes a size greater than 65535?

src/flexspi/nor.rs
felipebalbi: There's no `MAX_READ_SIZE`, do you mean `MAX_RECEIVE_SIZE`? That's 512. Anyway, you can receive 512 (or is it 1024?) bytes in one go. What happens when user requests 2048?

src/flexspi/nor.rs
```diff
@@ -987,9 +1014,7 @@ impl<'d> FlexspiNorStorageBus<'d, Blocking> {
             return Err(NorStorageBusError::StorageBusInternalError);
         }
 
-        for chunk in read_buf.chunks_mut(MAX_TRANSFER_SIZE as usize) {
-            self.read_cmd_data(chunk)?;
-        }
+        self.read_cmd_data(read_buf)?;
```
JamesHuard, line 1017: why can we not just do this in chunks over the MAX_RECEIVE_SIZE? I'm not understanding why we're returning an error for writes > 1024 and reads > 512

felipebalbi (changes requested):

src/flexspi/nor.rs
Shaibal-Microsoft: makes sense. I can add the check in the beginning

src/flexspi/nor.rs
felipebalbi: > > > What you should be doing is writing to the FIFO until it becomes full, then block until there's space in the FIFO for you to write more data (change the argument accordingly for the receive path).
> 
> And how do you detect FIFO is full? I don't see any interrupt for FIFO full.

you're not using interrupts, this is all blocking mode. Check the register prior to writing more.

> > > That's, essentially, what the DMA block would be doing for you, with the added limitation that DMA can only do 1024 bytes at a time, so you either chain transfer descriptors (which we don't, yet, support) or you handle the transaction in chunks of 1024 regardless of its total size.
> 
> In Blocking mode, why would you use DMA? Don't understand what you mean by this.

The "DMA" comment was to serve as an example.

PR #423, opened 2025-05-09
Move cargo-hack to a script

felipebalbi:
CI was starting to limit the amount of variations we could run in cargo hack. Instead, let's make a script that runs within CI and relies on cargo-batch to run all variations.

While at that, also make sure to add RTC timer as a variant.


felipebalbi:
Forgot to make sure CI installs thumbv8 target. I'll fix it shortly.

madeleyneVaca (approved):

PR #424, opened 2025-05-09
Make sure we're not using the PLL as main clock when resetting it.

de-vri-es:
This PR resets the main clock source to FFRO divided by 4 before changing the PLL configuration.

When a bootloader and payload application both use `embassy-imxrt`, the clock configuration routine is ran twice. The second time, the PLL will be reset while it's also being used as a main clock source. The result is that the CPU stops executing instructions and even the debug probe gets disconnected (which made this a bit difficult to debug :sweat_smile:).

By resetting the main clock source to the reset default, we can safely re-configure the PLL, and then switch (back) to it.

### DIFF
--- a/src/clocks.rs
+++ b/src/clocks.rs
@@ -937,6 +937,15 @@ impl MainPllClkConfig {
 }
 
 impl MainClkConfig {
+    /// Configure the FFRO/4 as the main clock source.
+    ///
+    /// This is the same as the reset value.
+    fn reset_main_clk() {
+        let clkctl0 = unsafe { crate::pac::Clkctl0::steal() };
+        clkctl0.mainclksela().write(|w| w.sel().ffro_div_4());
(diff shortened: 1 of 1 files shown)


felipebalbi:
> This PR resets the main clock source to FFRO divided by 4 before changing the PLL configuration.
> 
> When a bootloader and payload application both use `embassy-imxrt`, the clock configuration routine is ran twice. The second time, the PLL will be reset while it's also being used as a main clock source. The result is that the CPU stops executing instructions and even the debug probe gets disconnected (which made this a bit difficult to debug 😅).
> 
> By resetting the main clock source to the reset default, we can safely re-configure the PLL, and then switch (back) to it.

nice catch!!

felipebalbi (approved):

jerrysxie (approved):

PR #426, opened 2025-05-12
Change port events to use OOB and peripheral channel

philgweber:
Update example application and base eSPI code to use Peripheral and OOB based events rather than port based.

Both OOB and Peripheral events now contain the following details

/// Port event data
pub struct PortEvent {
    /// Port that event occurred on
    pub port: usize,

    /// Base address of buffer
    pub base_addr: u32,

    /// Offset into buffer
    pub offset: usize,

    /// Size of access
    pub length: usize,

    /// Direction of access
    pub direction: bool,
}

rt633 example code shows how to read these events. 

Added accessor method oob_get_write_buffer to return an unsafe handle to our OOB buffer in memory to write OOB data to. For all reads the base_addr is now returned as part of the notification event.

Changed name to oob_write_data to change wording to MCU side and assert if length > 0 as the length written to the HW buffers is len - 1 so a 0 length packet is illegal.


examples/rt633/src/bin/espi.rs
```diff
@@ -42,23 +43,26 @@ async fn main(_spawner: Spawner) {
             caps: Capabilities {
                 max_speed: Maxspd::SmallThan20m,
                 alert_as_a_pin: true,
+                allow_oob: true,
+                allow_128b_payload: true,
                 ..Default::default()
             },
             ram_base: 0x2000_0000,
-            base0_addr: 0x2002_0000,
```
philgweber, line 48: These parameters are optional now and not used by default so don't define them unless we are using base0 or base1.

src/espi.rs
```diff
@@ -623,65 +631,78 @@ impl<'d> Espi<'d> {
         });
     }
 
+    fn get_port_event(&mut self, port: usize) -> Poll<Result<Event>> {
```
philgweber, line 634: Same code for all ports, keep this code generic regardless of which port the event occurred on.

src/espi.rs
```diff
@@ -814,16 +835,33 @@ impl<'d> Espi<'d> {
         self.block_for_vwire_done();
     }
 
-    /// Trigger OOB read from host
+    /// Return pointer to OOB buffer buffer
+    ///
+    /// Warning: Writes to this buffer are unsafe as it is a direct mapped memory address
+    /// The length specified in config and location must be carved out in memory.x
+    pub unsafe fn oob_get_write_buffer(&mut self, port: usize) -> Result<&mut [u8]> {
```
philgweber, line 842: Not sure if we need further SAFETY warning here for accessing this buffer

src/espi.rs
```diff
@@ -814,16 +835,33 @@ impl<'d> Espi<'d> {
...
+            _ => Err(Error::InvalidPort),
+        }
+    }
+
+    /// Write OOB data from device to host
     /// When this completes it will send a INTWR event on OOB port
     /// Maximum len is 64
-    pub fn oob_read_start(&mut self, port: usize, length: u8) {
+    pub fn oob_write_data(&mut self, port: usize, length: u8) {
+        assert!(length > 0);
```
philgweber, line 858: I have assert to check length but if better I could return error code here if length == 0.

philgweber:
Closed by commit https://github.com/OpenDevicePartnership/embedded-services/commit/6ce0c0fb8813b81fb70cc1f4180a681a79bef844

examples/rt633/src/bin/espi.rs
```diff
@@ -42,23 +43,26 @@ async fn main(_spawner: Spawner) {
...
-            base1_addr: 0x2003_0000,
             status_addr: Some(0x480),
             status_base: Base::OffsetFrom0,
             ports_config: [
                 PortConfig::MailboxSplit {
                     direction: Direction::BidirectionalUnenforced,
-                    addr: 0,
+                    basesel: BaseOrAsz::OffsetFrom0,
```
felipebalbi, line 56: maybe `base_sel` for readability?

src/espi.rs
```diff
@@ -814,16 +835,33 @@ impl<'d> Espi<'d> {
         self.block_for_vwire_done();
     }
 
-    /// Trigger OOB read from host
+    /// Return pointer to OOB buffer buffer
```
felipebalbi, line 838: `s/buffer buffer/buffer/` ?

src/espi.rs
felipebalbi: yes. When is the buffer valid? What are the preconditions that must be maintained for this access to be safe?

src/espi.rs
felipebalbi: Produce an error, please. An assert will just `panic!()`. We want the user to have an opportunity to fix things.

src/espi.rs
```diff
@@ -994,51 +1032,15 @@ impl<'d> Espi<'d> {
...
-        // Enable the port
-        self.info.regs.mctrl().modify(|_, w| w.pena(port as u8).enabled());
-
-        // write 0x44 to data out
-        self.info
-            .regs
-            .port(port)
-            .dataout()
-            .write(|w| unsafe { w.data().bits(0x44) });
-    }
```
felipebalbi, line 1039: Why remove this?

src/espi.rs
```diff
@@ -113,10 +120,10 @@ pub enum PortConfig {
         /// Port Direction
         direction: Direction,
 
-        /// Port address to the host
-        addr: u16,
+        /// Base memory to select for port base
+        basesel: BaseOrAsz,
 
-        /// Offset into RAM space
+        /// 12-bit Word-aligned offset from selected base
```
felipebalbi, line 126: do you expect user to give you only the top 12 bits or the full 16 bits? It seems a bit unclear now with the change to rustdoc.

src/espi.rs
philgweber: For ACPI endpoint was there a reason to write 0x44 to DATAOUT? I didn't see any documentation on why we would be doing that so folded this in with the general mailbox configuration code unless there is something specific we need to set here?

src/espi.rs
philgweber: This goes back to the NXP docs which are a bit fuzzy on behavior here as there are two different registers to define offset. 

RAMUSE has 12-bit offset within the 4K window and can be used for any type of channel.
ADDR has 16-bit offset to select any offset but is only valid for Memory/IO and won't work for OOB.

If you define an offset in both RAMUSE and ADDR for a peripheral channel it's unclear what it actually does. As such the thing that makes most sense is to always use RAMUSE for both Peripheral and OOB which I've confirmed is working. However, this is limited to 12-bits so your offset must always be within the 4K region of RAMBASE/BASE0/BASE1.

If you want me to leave full flexibility I can allow the config for Peripheral to specify 2 different offsets, but then this can lead to issues when calculating physical location of data in address and specifying multiple offsets. As such I had removed access to setting offset in ADDR unless there is some use case we need this.

src/espi.rs
philgweber: Updated header with SAFETY details surrounding access to buffer region.

src/espi.rs
philgweber: Updated function to return Result and validate length range is 1..73

src/espi.rs
philgweber: Fixed comment.

src/espi.rs
felipebalbi: I copied that from NXP's samples. I think it's okay to remove, no worries.

RobertZ2011 (dismissed):

felipebalbi:
PR #428 vets `mimxrt633s-pac` v0.4.1

src/espi.rs
```diff
@@ -623,65 +634,78 @@ impl<'d> Espi<'d> {
...
+        }
+
+        // Common registers for all ports
+        let datain = self.info.regs.port(port).datain().read();
+        let idxoff = datain.idx().bits() as usize;
+        let length = datain.data_len().bits() as usize + 1;
+        let direction = datain.dir().bit_is_set();
+
+        // By default set base address to RAMBASE
+        let mut address = self.config.ram_base as u32;
```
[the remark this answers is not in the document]
philgweber: Updated thanks this is cleaner.

felipebalbi (approved):

jerrysxie (approved):

PR #430, opened 2025-05-16
Improve teaching value of examples

de-vri-es:
Hey,

This PR makes a few adjustments to the examples in an attempt to improve their teaching value. The examples will probably be an important tool for new users, so I think it is good to try and polish them a little more. They're already quite useful, but I did notice a few things that could be improved :)

The PR:
1) Replaces `extern crate embassy_imxrt_examples` with `use embassy_imxrt_examples as _`. This is the standard way it is done for external crates that need to be linked in, like `defmt_rtt` and `panic_probe`.
2) Replaces star imports with specific imports. This makes it more explicit what is actually imported, which helps to read the code. In general, star imports are usually frowned upon (with the exception of `pub use self::submodule::*`).
3) Simplifies the DMA example by removing the custom macro.
4) Fixes the trait sealing in the storage service example (it did not actually seal the trait).
5) Moves the `defmt_rtt` and `panic_probe` imports out of the library and into specific examples. If users use the examples as a starting point it would be best if the example is pretty much complete without hiding details in a custom library.

    Arguably, this also means that the `BIV`, `OTFAD`, `FCB` and `KEYSTORE` should be moved to the specific examples. But that's a bit more duplication than I wanted to do in this PR without discussion first.

They're all separate commits, so it's relatively straightforward to remove any of them if you do not agree with (all) the changes.

### DIFF
--- a/examples/rt633/src/bin/espi.rs
+++ b/examples/rt633/src/bin/espi.rs
@@ -1,8 +1,6 @@
 #![no_std]
 #![no_main]
 
-use rt633_examples as _;
-
 use core::slice;
 
 use defmt::{error, info};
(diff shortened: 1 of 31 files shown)


felipebalbi (dismissed):
Thank you, very nice cleanups 👍

felipebalbi (approved):
Updated format

RobertZ2011 (approved):

PR #431, opened 2025-05-20
timer: ostimer: wait for match write ready

felipebalbi:
Before writing to MATCH_L/MATCH_H registers, wait for permission to do so by pollinmg MATCH_WR_RDY bit in CTRL register.

### DIFF
--- a/src/time_driver/ostimer.rs
+++ b/src/time_driver/ostimer.rs
@@ -84,6 +84,10 @@ impl OsTimer {
         let alarm = self.alarms.borrow(cs);
         alarm.timestamp.set(timestamp);
 
+        // Wait until we're allowed to write to MATCH_L/MATCH_H
+        // registers
+        while os().osevent_ctrl().read().match_wr_rdy().bit_is_set() {}
+
         let t = self.now();
         if timestamp <= t {
             os().osevent_ctrl().modify(|_, w| w.ostimer_intena().clear_bit());
@@ -114,7 +118,8 @@ impl OsTimer {
     fn on_interrupt(&self) {
         critical_section::with(|cs| {
             if os().osevent_ctrl().read().ostimer_intrflag().bit_is_set() {
-                os().osevent_ctrl().modify(|_, w| w.ostimer_intena().clear_bit());
+                os().osevent_ctrl()
+                    .modify(|_, w| w.ostimer_intena().clear_bit().ostimer_intrflag().set_bit());
                 self.trigger_alarm(cs);
             }
         });


src/time_driver/ostimer.rs
```diff
@@ -84,6 +84,10 @@ impl OsTimer {
         let alarm = self.alarms.borrow(cs);
         alarm.timestamp.set(timestamp);
 
+        // Wait until we're allowed to write to MATCH_L/MATCH_H
+        // registers
+        while os().osevent_ctrl().read().match_wr_rdy().bit_is_set() {}
```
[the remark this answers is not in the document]
jerrysxie: Unlikely to happen since the delay should never be more than 3 bus clocks.

src/time_driver/ostimer.rs
```diff
@@ -114,7 +118,8 @@ impl OsTimer {
     fn on_interrupt(&self) {
         critical_section::with(|cs| {
             if os().osevent_ctrl().read().ostimer_intrflag().bit_is_set() {
-                os().osevent_ctrl().modify(|_, w| w.ostimer_intena().clear_bit());
+                os().osevent_ctrl()
```
jerrysxie, line 121: Did adding this change any manifested behavior? I didn't see any change adding this.

src/time_driver/ostimer.rs
```diff
@@ -84,6 +84,10 @@ impl OsTimer {
         let alarm = self.alarms.borrow(cs);
         alarm.timestamp.set(timestamp);
 
+        // Wait until we're allowed to write to MATCH_L/MATCH_H
```
jerrysxie, line 87: Should we extract this out into a separate function so we always check this before writing to match because we also write to match in `init()`. We should just test this bit to makes we never try to load it without it being ready to accept.

jerrysxie (changes requested):

src/time_driver/ostimer.rs
felipebalbi: I didn't see any noticeable change either, but the documentation does call it out.

src/time_driver/ostimer.rs
felipebalbi: when the HW starts, this bit is cleared :)

felipebalbi: This has been running for well over 1 hr without any lock-ups. Pretty sure we need both bits set.

jerrysxie (approved):

tullom (approved):

PR #433, opened 2025-05-20
Calibrate delay loop clocks

felipebalbi:
Calibrate the divisor to make sure we get as close as possible to the
requested delay.

While at that, also rename the frequency argument to `freq_hz` as
that's what the method expects. Additionally, update rustdoc to
further explain how the method is supposed to work.

### DIFF
--- a/src/clocks.rs
+++ b/src/clocks.rs
@@ -1282,8 +1282,23 @@ impl ConfigurableClock for SysOscConfig {
 }
 
 /// Method to delay for a certain number of microseconds given a clock rate
-pub fn delay_loop_clocks(usec: u64, freq_mhz: u64) {
-    let mut ticks = usec * freq_mhz / 1_000_000 / 4;
+///
+/// Given `usec` and `freq_hz`, this method will compute the number of
+/// ticks to be passed to `cortex_m::asm::delay()` such that we reach
+/// the amount of microseconds requested by the caller.
+pub fn delay_loop_clocks(usec: u64, freq_hz: u64) {
+    // NOTICE: The correct math would be:
+    //
+    //     usec * 1_000 / 1_000_000_000 / freq_hz
+    //
+    // Which simplifies to:
+    //
+    //     usec * freq_hz / 1_000_000;
+    //
+    // However, testing shows that we're always about 50% over the
+    // requested target. Adding that extra 50% to the divisor gets us
+    // very close to what was requested.
+    let mut ticks = usec * freq_hz / 1_500_000;
     if ticks > u64::from(u32::MAX) {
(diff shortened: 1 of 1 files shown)


src/clocks.rs
```diff
@@ -1282,8 +1282,23 @@ impl ConfigurableClock for SysOscConfig {
...
+pub fn delay_loop_clocks(usec: u64, freq_hz: u64) {
+    // NOTICE: The correct math would be:
+    //
+    //     usec * 1_000 / 1_000_000_000 / freq_hz
+    //
+    // Which simplifies to:
+    //
+    //     usec * freq_hz / 1_000_000;
+    //
+    // However, testing shows that we're always about 50% over the
```
RobertZ2011, line 1298: Do we have any idea what could be causing the extra 50%? This makes me slightly worried that we're configuring our clock incorrectly somewhere.

RobertZ2011 (approved):

src/clocks.rs
tullom: Same question, will approve once this is addressed

felipebalbi: We're not. It's very likely a side effect of M33's pipeline. That is, M33, on average, executes 1.5 instructions per cycle.

You can look at the generated assembly:

image

And the source from `cortex-m` crate, which contains some nice explanation of the thing:

https://github.com/rust-embedded/cortex-m/blob/master/cortex-m/src/asm.rs#L33-L53

Moreover, Cortex-M33 has a 3-stage in-order pipeline. Assuming no hazards of any kind, at every cycle we have one instruction completing, meaning that we should have the loop complete within `2` cycles. That, in turn, means that, in an ideal scenario, we should *not* have that extra 50% of overhead; however, this MCU does not have a branch a predictor, what I'm assuming (which I'm yet to confirm because I couldn't find cycle count for M33 instructions anywhere) is that branching takes twice as long as that `SUBS`, this resulting in the loop taking `3` cycles, instead of two.

felipebalbi: Yup, pretty sure my hunch is correct. Here's a sample using DWT:

image

And the result is:

```console
2175.152967 [INFO ] Hello world (hello_world src/bin/hello-world.rs:13)
2175.159077 [INFO ] DWT Cycle Counter That took 1500081 cycles!!! (hello_world src/bin/hello-world.rs:25)
```

Pretty close to 1.5M exactly.

tullom: Very good sleuthing Detective Balbi 🕵️

tullom (approved):

felipebalbi:
Okay, final comment:

After the first iteration, it becomes very predictable:

image

PR #439, opened 2025-05-25
Use unused LUT slot in FlexSPI storage bus driver.

de-vri-es:
The storage bus driver for FlexSPI currently uses look-up-table index 0. However, this index is also used by the ROM bootloader, and will still be used by the CPU to access flash through the memory bus.

A simple solution is to use index 14 instead, as this is not used in the default configuration.

image

"Reserved" in the table means free to use ^_^

Note that the `mimxrt600-fcb` crate uses index 6 to configure the device in DDR mode, so that one is also already taken.

But a user might be on a platform that overrides which command sequence is used for what purpose, so the PR also allows the user to choose their own.

This makes it a much easier to use the FlexSpi storage bus driver while keeping the configuration set by the ROM bootloader from the FCB.

### DIFF
--- a/src/flexspi/nor.rs
+++ b/src/flexspi/nor.rs
@@ -93,7 +93,7 @@ macro_rules! configure_ports_b {
 
 const FIFO_SLOT_SIZE: u32 = 4; // 4 bytes
 const MAX_TRANSFER_SIZE: u32 = 128;
-const OPERATION_SEQ_NUMBER: u8 = 0;
+const OPERATION_SEQ_NUMBER: u8 = 14;
 const LUT_UNLOCK_CODE: u32 = 0x5AF05AF0;
 
 #[cfg(feature = "time")]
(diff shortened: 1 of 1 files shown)


src/flexspi/nor.rs
```diff
@@ -635,6 +641,22 @@ impl<'d> BlockingNorStorageBusDriver for FlexspiNorStorageBus<'d, Blocking> {
...
 
 impl<'d, M: Mode> FlexspiNorStorageBus<'d, M> {
+    /// Set the command sequence in the FlexSPI LUT to use.
+    ///
+    /// All commands sent over the FlexSPI bus are first programmed into a lookup-table.
+    /// You should make sure that the driver uses a command sequence that is not already used for other purposes.
+    ///
+    /// By default, we use command sequence 14, because it not used by the bootloader ROM or the `mimxrt600_fcb` crate (in the default configuration).
+    /// However, if you are on a platform where sequence 14 is already in use, you should select a different one.
+    pub fn set_command_sequence_number(&mut self, value: u8) {
```
RobertZ2011, line 651: Is this something that can be passed as a configuration option, maybe through something like `FlexspiConfigPortData`? This function would allow changing this configuration option anytime during runtime. Is that something that's safe to do/something we want to allow?

PR #443, opened 2025-06-01
Fix warning in keyboard example.

de-vri-es:
The `extern crate embassy_imxrt_examples;` and `use embassy_imxrt_examples as _` have the same effect. Only one is enough, and the second one is more idiomatic :)


de-vri-es:
Ah, nice. Didn't see that the fix was already there!

PR #447, opened 2025-06-11
move systemview-tracing cfg out of macro and into helper function

kat-perez:
- Refactored interrupt handler macro to delegate to a helper function (`__handle_interrupt`) for conditional tracing.
- Moved all `#[cfg(feature = "systemview-tracing")]` logic out of the macro and into the helper function, ensuring correct evaluation in the defining crate.
- Added required trait bound `T: crate::interrupt::typelevel::Interrupt` to `__handle_interrupt` for trait compliance.
- This eliminates downstream cfg warnings and keeps macro-generated code clean and portable.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -65,6 +65,27 @@ pub use embassy_hal_internal::{Peri, PeripheralType};
 #[cfg(feature = "rt")]
 pub use crate::pac::NVIC_PRIO_BITS;
 
+// Helper function for interrupt handling with optional tracing
+#[inline(always)]
+#[doc(hidden)]
+pub unsafe fn __handle_interrupt<T, H>()
+where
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -65,6 +65,27 @@ pub use embassy_hal_internal::{Peri, PeripheralType};
 #[cfg(feature = "rt")]
 pub use crate::pac::NVIC_PRIO_BITS;
 
+// Helper function for interrupt handling with optional tracing
+#[inline(always)]
+#[doc(hidden)]
+pub unsafe fn __handle_interrupt<T, H>()
```
felipebalbi, line 71: you should probably mark this `#[inline(always)]`

RobertZ2011 (approved):

PR #449, opened 2025-06-18
Fix clippy warnings

de-vri-es:
This PR fixes most clippy warnings.

Three warnings remain because they would require changing the public API. They are all in the flexspi nor storage bus module: two cases of returning `Result<_, ()>` in a public API and one case of too many arguments in a public function.


src/flexspi/nor.rs
```diff
@@ -965,14 +958,7 @@ impl<'d, M: Mode> FlexspiNorStorageBus<'d, M> {
...
 
-        match cmd.dummy {
-            NorStorageDummyCycles::Clocks(clk) => {
-                if clk > 0 {
-                    self.program_dummy_instruction(cmd, &mut cookie);
-                }
-            }
-            _ => {}
-        }
+        self.program_dummy_instruction_if_non_zero(cmd, &mut cookie);
```
de-vri-es, line 961: This actually changes behavior: previously, all NorStorageDummyCycles::Bytes were ignored. Now they are passed to `program_dummy_instruction`. Although I have my doubts about the conversion in that function that simply equates bytes to clock cycles.

kurtjd (approved):

src/flexspi/nor.rs
kurtjd: Can disregard if the change in behavior is expected.

PR #461, opened 2025-07-31
Feed watchdog immediately after unleashing it

### LINKED ISSUE #460: watchdog needs to be fed once before it is fully enabled
Image

Image

We should feed the watchdog once in `unleash()` once because that would align with expectations.

kurtjd:
Watchdog must be fed after enabling, so call `feed()` within `unleash()`. This was done originally but a commit accidentally removed this functionality while making other changes, so reintroduce that line.

 Resolves #460


jerrysxie (approved):

felipebalbi (approved):
Probably a good idea to reference the datasheet section where this is stated.

PR #463, opened 2025-07-31
Add i2c-master-drop example

jerrysxie:
Add an example showing how to pass pins into an I2C peripheral, then
drop and recreate the I2C peripheral

Based on #462

### DIFF
--- a/examples/rt685s-evk/src/bin/i2c-master-drop.rs
+++ b/examples/rt685s-evk/src/bin/i2c-master-drop.rs
@@ -1,0 +1,91 @@
+#![no_std]
+#![no_main]
+
+use defmt::{error, info};
+use embassy_executor::Spawner;
+use embassy_imxrt::{bind_interrupts, i2c, peripherals};
+use embassy_time::Timer;
+use embedded_hal_async::i2c::I2c;
+use {defmt_rtt as _, embassy_imxrt_examples as _, panic_probe as _};
+
+const ACC_ADDR: u8 = 0x1E;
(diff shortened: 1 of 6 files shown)


kurtjd (approved):

examples/rt685s-evk/src/bin/i2c-master-drop.rs
```diff
@@ -0,0 +1,91 @@
...
+    info!("i2c example - I2cMaster dropped");
+
+    {
+        info!("i2c example - I2cMaster::new after drop");
+        let mut i2c = i2c::master::I2cMaster::new_async(
+            p.FLEXCOMM2.reborrow(),
+            p.PIO0_18.reborrow(),
+            p.PIO0_17.reborrow(),
+            Irqs,
+            Default::default(),
```
felipebalbi, line 71: I suggest changing the configuration to something else. Maybe change bus frequency. This will help make sure that dropping and reinitializing does not leave bogus configs behind. Could be done as a separate PR too.

felipebalbi (approved):

examples/rt685s-evk/src/bin/i2c-master-drop.rs
jerrysxie: Done, reconfigure the i2c bus timing after drop, and verify on the scope that the new configuration took effect as expected.

felipebalbi (approved):

kurtjd (approved):

PR #468, opened 2025-08-09
flexspi: INTR should not do read-modify-write

jerrysxie:
All bitfields in this INTR are write 1 to clear, read-modify-write will accidentally clear other bits that are set.


williampMSFT (dismissed):
:shipit:

kurtjd (approved):

PR #471, opened 2025-08-19
rng: add SW entropy test based on NIST SP 800-90B

jerrysxie:
Since HW entropy test has HW bug when main clock is running at high clock (according to NXP), add SW entropy test based on NIST SP800-90B.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -50,7 +50,7 @@ jobs:
         run: cargo fmt --check
         working-directory: ${{ matrix.workdir }}
 
-  clippy:
+  clippy-examples:
     runs-on: ubuntu-latest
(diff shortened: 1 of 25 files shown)


src/rng.rs
```diff
@@ -26,6 +32,9 @@ pub enum Error {
 
     /// Frequency Count Fail
     FreqCountFail,
+
+    /// Entropy Error.
+    EntropyError,
```
jerrysxie, line 37: @felipebalbi Is `SeedError` meant to be the same idea as `EntropyError`?

jerrysxie: Gotcha, consolidate the error into one.

kurtjd (approved):

PR #472, opened 2025-08-19
espi: remove unneeded async from complete_port()

jerrysxie:


kurtjd (approved):
I believe this is a breaking change but otherwise looks good

jerrysxie:
> I believe this is a breaking change but otherwise looks good

Good point 👍

jerrysxie:
> > I believe this is a breaking change but otherwise looks good
> 
> Good point 👍

Announced on Zulip: [#embedded-controller > embassy-imxrt espi API breaking changes](https://opendevicepartnership.zulipchat.com/#narrow/channel/495405-embedded-controller/topic/embassy-imxrt.20espi.20API.20breaking.20changes)

Issue #477, opened 2025-08-25
Clock tree configuration is not used


Wassasin:
Currently the clock tree configuration type definition does not reflect the current level of implementation within the HAL. For example, setting any value like clock source or frequency for the main clock is not used at all.

For example `self` is not used at all: https://github.com/OpenDevicePartnership/embassy-imxrt/blob/642d05ff8b53d772a0613bcf5f10d6e7a703b178/src/clocks.rs#L1098-L1101

Secondly, some fields within the configuration do not actually reflect configuration parameters, but instead might be set after the fact to record the frequency for the clock domain. Instead of using the actual register values or original configuration to compute this, a hard coded configuration block is used instead.

For example, here the default `crystal` configuration is hardcoded: https://github.com/OpenDevicePartnership/embassy-imxrt/blob/642d05ff8b53d772a0613bcf5f10d6e7a703b178/src/timer.rs#L369

Here `State` denoting `enabled/disabled` is used for the trait `ConfigurableClock` but does not actually reflect the current state. Some clock domains have this not implemented: https://github.com/OpenDevicePartnership/embassy-imxrt/blob/642d05ff8b53d772a0613bcf5f10d6e7a703b178/src/clocks.rs#L403-L405

Either this should be fully implemented, or any configuration value for which setting it is not yet configured should be (temporarily) removed, in order to make clear what is actually working and what not.

PR #478, opened 2025-08-25
Added mainclk source selection

Wassasin:
When implementating the bootloader, walked head-first into the problem that the clock domain configuration is partially implemented. See https://github.com/OpenDevicePartnership/embassy-imxrt/issues/477 .

For now, I quickly added setting the mainclk source. For the bootloader I require a slower main clock. This PR should not change the default behaviour.

### DIFF
--- a/src/clocks.rs
+++ b/src/clocks.rs
@@ -95,9 +95,8 @@ impl ClockConfig {
             hclk: HclkConfig { state: State::Disabled },
             main_clk: MainClkConfig {
                 state: State::Enabled,
-                //FFRO divided by 4 is reset values of Main Clk Sel A, Sel B
                 src: MainClkSrc::PllMain,
-                div_int: AtomicU32::new(4),
+                div_int: AtomicU32::new(2),
                 freq: AtomicU32::new(CORE_CPU_FREQ),
             },
             main_pll_clk: MainPllClkConfig {
@@ -1341,12 +1340,13 @@ fn set_pad_voltage_range() {
 }
 
 /// Initialize AHB clock
-fn init_syscpuahb_clk() {
+fn init_syscpuahb_clk(divisor: u16) {
     // SAFETY: unsafe needed to take pointer to Clkctl0
     let clkctl0 = unsafe { crate::pac::Clkctl0::steal() };
     // SAFETY: unsafe needed to write the bits
-    // Set syscpuahbclkdiv to value 2, Subtract 1 since 0-> 1, 1-> 2, etc...
-    clkctl0.syscpuahbclkdiv().write(|w| unsafe { w.div().bits(2 - 1) });
+    clkctl0
+        .syscpuahbclkdiv()
+        .write(|w| unsafe { w.div().bits(divisor.saturating_sub(1) as u8) });
 
     while clkctl0.syscpuahbclkdiv().read().reqflag().bit_is_set() {}
 }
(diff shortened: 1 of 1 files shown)


JamesHuard (dismissed):

src/clocks.rs
```diff
@@ -1513,9 +1530,10 @@ fn init_clock_hw(config: ClockConfig) -> Result<(), ClockError> {
         cc0.espiclksel().write(|w| w.sel().use_48_60m());
     }
 
+    config.main_clk.enable_and_reset()?;
```
Wassasin, line 1533: Also fixed ordering of this call as it should be configured before starting syscpuahb_clk.

src/clocks.rs
```diff
@@ -946,13 +946,30 @@ impl MainClkConfig {
...
-    fn init_main_clk() {
+    fn init_main_clk(&self) {
         // SAFETY:: unsafe needed to take pointers to Clkctl0 and Clkctl1
         // used to set the right HW frequency
         let clkctl0 = unsafe { crate::pac::Clkctl0::steal() };
         let clkctl1 = unsafe { crate::pac::Clkctl1::steal() };
 
-        clkctl0.mainclkselb().write(|w| w.sel().main_pll_clk());
+        let (clk_a, clk_b) = {
+            use mimxrt685s_pac::clkctl0::mainclksela::Sel as SelA;
```
jerrysxie, line 956: This should be `crate::pac` because it can be a different PAC based on feature flags.

jerrysxie (changes requested):

src/clocks.rs
Wassasin: Actually, could you refer me to the note that the divider must be set before setting the source clock? The note from UM11147 does not actually specify that the source clock can be changed after setting the divider.

src/clocks.rs
Wassasin: Fixed!

src/clocks.rs
```diff
@@ -946,13 +946,30 @@ impl MainClkConfig {
...
+                MainClkSrc::SFRO => (None, SelB::SfroClk),
+                MainClkSrc::PllMain => (None, SelB::MainPllClk),
+                MainClkSrc::RTC32k => (None, SelB::Rtc32kClk),
+            }
+        };
+
+        if let Some(clk_a) = clk_a {
+            clkctl0.mainclksela().write(|w| w.sel().variant(clk_a));
+        }
+        clkctl0.mainclkselb().write(|w| w.sel().variant(clk_b));
```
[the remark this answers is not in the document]
jerrysxie: `MultiSourceClock::set_clock_source_and_rate()` is never called by anyone. I think there was some more ambitious idea for the clock tree initially, but it was only half done.

While the default for config changes, it actually matches the behavior now, so I wouldn't consider that to be breaking change, more like a bug on `embassy-imxrt` part. It should have no effect on clients, because client as use the default as far as I can tell.

src/clocks.rs
jerrysxie: > Have we reached out to NXP on why we might see lock ups in one order versus the other, and if they have a clear response on which is "correct" in a given scenario?

Yeah, they told us to do what the SDK does and did not provide a lot more than that.

jerrysxie: > Added some code to first set the divider to 0xff, before switching over to another clock, and setting the final divider.

That makes sense, and it works on my board. If we switched over to MainPLL while the divider is at default value of 1, the sys clock would be running at 500 MHz . That probably is what is causing the CPU lockup. Doing it this way, will provide the clock rate to be too fast if we are trying to speed up the clock.

jerrysxie (approved):

JamesHuard (approved):

PR #492, opened 2025-09-16
Async passthough

diondokter:
With some refactoring we can be smarter than the compiler (which isn't too smart when it comes to async).
This saves some flash!

I don't know how much in a *real* application, but all examples combined save ~4.5kB.

Al behavior and public APIs should still be the same except for the one in dma/channel.rs, but that one will be an obvious refactor if that breaks anyone.

### DIFF
--- a/src/dma/channel.rs
+++ b/src/dma/channel.rs
@@ -1,8 +1,6 @@
 //! DMA channel & request
 
-use core::future::poll_fn;
 use core::marker::PhantomData;
-use core::task::Poll;
 
 use embassy_sync::waitqueue::AtomicWaker;
 
@@ -35,15 +33,13 @@ impl<'d> Channel<'d> {
     }
 
     /// Writes from a memory buffer to another memory buffer
-    pub async fn write_to_memory(
+    pub fn write_to_memory(
         &'d self,
         src_buf: &'d [u8],
         dst_buf: &'d mut [u8],
         options: TransferOptions,
     ) -> Transfer<'d> {
-        let transfer = Transfer::new_write_mem(self, src_buf, dst_buf, options);
-        self.poll_transfer_complete().await;
-        transfer
+        Transfer::new_write_mem(self, src_buf, dst_buf, options)
     }
 
     /// Return a reference to the channel's waker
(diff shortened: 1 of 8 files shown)


JamesHuard (approved):

src/i2c/master.rs
```diff
@@ -963,7 +963,7 @@ impl<'a> I2cMaster<'a, Async> {
         }
     }
 
-    async fn stop(&mut self) -> Result<()> {
+    fn stop(&mut self) -> Result<impl Future<Output = Result<()>> + use<'a, '_>> {
```
kurtjd, line 966: I was working on implementing this optimization on a different codebase and was considering cases like this where there is some sync code followed by a single await point.

I think in cases like this there is technically a change in behavior. By marking the function async, the entire block won't run until awaited. By doing this passthrough, all the sync code will run immediately even before the function is awaited. I don't think we really rely on this difference in behavior anywhere, though do you think it's something to be concerned about? If not, I can apply this to a few more places where I was working.

diondokter: Hmmm, yeah that is true. Though I find I never create a future only to then never poll it expecting nothing will have happened. In theory that is behavior you may wish exists, but it's never come up for me.

So calling this function and then dropping the future is the same as polling the old function once and then dropping. In practice I don't think this is going to cause any problems.

kurtjd: Sounds good, yeah didn't think this would really cause too many problems in practice so glad you confirmed it.

src/dma/channel.rs
```diff
@@ -35,15 +33,13 @@ impl<'d> Channel<'d> {
...
+    pub fn write_to_memory(
         &'d self,
         src_buf: &'d [u8],
         dst_buf: &'d mut [u8],
         options: TransferOptions,
     ) -> Transfer<'d> {
-        let transfer = Transfer::new_write_mem(self, src_buf, dst_buf, options);
-        self.poll_transfer_complete().await;
-        transfer
+        Transfer::new_write_mem(self, src_buf, dst_buf, options)
```
jerrysxie, line 42: This is a pretty big behavior change. Now `write_to_memory()` no longer guarantee that the DMA transfer is done, and with the current API the DMA driver does not offer a public async API to wait for DMA to be completed. It seems like a bigger refactoring will be needed for this approach to work with the DMA driver. Maybe we can do undo this one for now.

jerrysxie (changes requested):

felipebalbi (approved):

PR #494, opened 2025-09-21
rng: set sdctl register to recommended values

jerrysxie:
With these recommended values, the entropy check failure rate is 2 out of 100,000. Both RCT failures.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -174,7 +174,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


felipebalbi:
> > Nit: maybe the conversion to `is_multiple_of()` could have been carried out as a separate PR.
> 
> 
> 
> CI pipeline fails due to the rustc 1.90 update with the original PR, but if we don't mind merging the original PR with failing CI, then We can split it in two PRs.

If there's an actual failure without it, then maybe ignore this nitpicking. Just merge it :-)

PR #500, opened 2025-10-15
Increment multiplier if i2c frequency exceeds spec.

GBJin:
We have observed that i2c frequency has possibility to overshoot the spec (101.01k Hz observed in 100k Hz standard mode).
Add mechanism to increment multiplier before returning final frequency allows actual frequency from scope lower than spec.
Frequency overshoot is not allowed according to spec UM10204, table.10
image

I2c frequency has been empirically found to be running faster.
Here's the waveform captured from scope, 101.01kHz is observed in standard mode (100k) with rust implementation.
image

Here's previous C implementation:
image

And current Rust implementation
image

### DIFF
--- a/src/i2c/master.rs
+++ b/src/i2c/master.rs
@@ -114,7 +114,7 @@ impl SpeedRegisterSettings {
         // Figure out what we need to set the clock divider to in order to hit the I2C speed the user requested.  Again, we may not
         // be able to be exact, so we need to find the closest viable option.
         //
-        let (result_clocks_hi, result_clocks_lo, result_div_multiplier) = (MIN_CLOCKS..=MAX_CLOCKS)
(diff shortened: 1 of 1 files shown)


felipebalbi:
> We have observed that i2c frequency has possibility to overshoot the spec (100.8k Hz observed in 100k Hz standard mode). 

What were the DIV, HIGH, and LOW values chosen when you noticed this problem?

felipebalbi:
This is fixed through #501 . @GBJin , can you verify that all tests pass with @williampMSFT's PR?

GBJin:
> This is fixed through #501 . @GBJin , can you verify that all tests pass with @williampMSFT's PR?

I'll provide test results next week with latest imxrt.

PR #501, opened 2025-10-16
I2C: Add option to slightly undershoot target frequency

williampMSFT:
In some cases, it looks like when a target frequency is chosen, the hardware will sometimes emit one or two waves that are slightly shorter than that target that would, if sustained, overshoot the target frequency.  We have not observed these shorter waves being sustained.

This change adds an optional flag to allow clients to request the I2C system to slightly underclock the bus.  This will provide headroom so that, in these cases, the slightly-shorter waves will still be long enough to not overshoot the target frequency.

### DIFF
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -791,8 +791,7 @@ macro_rules! impl_instance {
     }
 }
 
-// REVISIT: Add support for FLEXCOMM14 with high speed SPI.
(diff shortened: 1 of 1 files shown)


felipebalbi (approved):

felipebalbi:
Standard:

image


Fast:

image


Fast Plus:

image

PR #502, opened 2025-10-21
I2c: use lowest pre-divider possible

felipebalbi:
By using lowest possible pre-divider with highest possible `HIGH` and `LOW` counts, we achieve a much more stable I2C clock.

### DIFF
--- a/src/i2c/master.rs
+++ b/src/i2c/master.rs
@@ -126,7 +126,14 @@ impl SpeedRegisterSettings {
                 (hi_clocks, lo_clocks, clock_div_multiplier)
             })
             .filter(|(hi_clocks, lo_clocks, clock_div_multiplier)| {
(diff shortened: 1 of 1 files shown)


felipebalbi:
Running well after a few hours:

image

jerrysxie (approved):
Have we tested 400 kHz as well? How does it look at 400 kHz?

jerrysxie (commented):
Have we tested 400 kHz as well? How does it look at 400 kHz?

williampMSFT:
> Have we tested 400 kHz as well? How does it look at 400 kHz?
yeah, there's a little more overshoot there [edit: more by absolute value, not percentage]:
image

Chatted offline with Felipe, I think the plan is to get this in to mitigate a little bit and then rework https://github.com/OpenDevicePartnership/embassy-imxrt/pull/501 on top of this to add like a 'strict mode' that imposes like a 3% safety margin on the target speed

felipebalbi:
> Have we tested 400 kHz as well? How does it look at 400 kHz?

image

PR #503, opened 2025-10-22
Add FLEXCOMM14 support for spi 

KretonWang:
Add FLEXCOMM14 support for spi

### DIFF
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -791,7 +791,6 @@ macro_rules! impl_instance {
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/spi.rs
```diff
@@ -792,7 +792,7 @@ macro_rules! impl_instance {
 }
 
 // REVISIT: Add support for FLEXCOMM14 with high speed SPI.
```
kurtjd: Can this comment be removed now?

felipebalbi (approved):

PR #515, opened 2025-12-09
rng: remove panics

felipebalbi:
Fixes https://github.com/OpenDevicePartnership/embassy-imxrt/issues/520

### DIFF
--- a/src/i2c/slave.rs
+++ b/src/i2c/slave.rs
@@ -255,8 +255,8 @@ impl<'a> I2cSlave<'a, Async> {
     ) -> Result<Self> {
         let ch = dma::Dma::reserve_channel(dma_ch);
 
-        if ch.is_some() {
-            let this = Self::new_inner::<T>(_bus, scl, sda, address, Some(ch.unwrap()))?;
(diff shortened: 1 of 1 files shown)


jerrysxie:
@felipebalbi Can we remove the `RngCore` trait implementation? For random number generation which depends on real entropy source and can fail. The methods in this trait seems not suitable. Is there another Rng trait that is fallible and return results.

kurtjd (approved):

src/rng.rs
```diff
@@ -131,6 +144,53 @@ impl<'d> Rng<'d> {
...
+
+        if entropy.contains(&0) {
+            return Err(Error::SeedError);
+        }
+
+        sw_entropy_test(&entropy)?;
+
+        // SAFETY: entropy is the same for input and output types in
+        // native endianness.
+        let entropy: [u8; 64] = unsafe { core::mem::transmute(entropy) };
```
asasine: Since this relies on `entropy` being `[u32; 16]`, I think it would be helpful to mark that explicitly on the type instead of relying on the fact that integer literals default to `i32`

PR #522, opened 2025-12-10
i2c: fix master start bug in async write_read()

jerrysxie:
The flag passed to the second start is wrong, yikes! Fortunately client use this API through `embedded_hal_async` which just calls `transaction()`: https://docs.rs/embedded-hal-async/latest/src/embedded_hal_async/i2c.rs.html#121

Testing on 685 evk with these examples:
- i2c-master
- i2c-master-async
- i2c-master-drop
- i2c-loopback-async
- i2c-loopback-async-10bit


jerrysxie:
@felipebalbi @williampMSFT @jamesmunns Maybe we should drop the other trait method implementation and only implement `transaction()`?

felipebalbi (approved):

PR #525, opened 2025-12-12
Add support for RX DMA always enabled UART

Neil-alter:
I have observed a 16 bytes hang issue in UART reading. The root cause is that DMA is being enabled and disabled on every read() call, which leads to byte loss when the incoming packet is large, which leads to 16‑byte hang when the incoming packet is large.
During DeepSleep, the UART sometimes encounters a 16‑byte hang. To resolve this, the following method want to inject in:
1. Add read_buffered logic
This allows continuous data retrieval from DMA without toggling the DMA state for every read, preventing data loss on large packets.
2. Handle interrupts correctly
~~When enabling TXLVL and RXLVL in FIFOINTENSET to make FCWAKE in HWAKE effective, the corresponding TXLVL and RXLVL interrupts must be cleared in the interrupt handler to prevents an interrupt storm, and clear RXERR and TXERR bit if it is set.

This pull request adds comprehensive support for ping-pong (double buffer) DMA transfers, enabling more efficient and continuous data movement with overrun error detection and buffer management. The changes introduce new data structures, descriptor management, and APIs for configuring and monitoring ping-pong transfers, as well as enhancements to the DMA interrupt handler to support seamless buffer switching.

**Ping-pong DMA support and buffer management:**

* Added new static descriptor arrays (`PING_DESCRIPTORS`, `PONG_DESCRIPTORS`) and status tracking (`PING_PONG_STATUS`) to manage ping-pong buffer states for each DMA channel. Introduced `PingPongSelector` and `BufferStatus` enums, and the `PingPongStatus` struct to represent buffer selection and status.
* Implemented `configure_channel_ping_pong` in `Channel` to set up a DMA channel for ping-pong operation, initializing descriptors and buffer statuses for double-buffered transfers.
* Added methods to `Channel` for querying and managing ping-pong buffers: `current_buffer`, `commit_buffer` (unsafe, marks a buffer as ready for DMA), `buffer_status`, and `check_and_clear_overrun_error` for error handling.

**DMA interrupt handler enhancements:**

* Updated the DMA IRQ handler to automatically switch between ping and pong buffers upon transfer completion, update buffer statuses, and detect/flag buffer overrun errors if both buffers are not ready in time.

**API improvements:**

* Added `get_channel_number` to `Channel` for retrieving the DMA channel number, aiding in diagnostics and management.

**Imports and code organization:**

* Updated imports in `channel.rs` to include new types and statics for ping-pong support.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -414,7 +414,7 @@ dependencies = [
 [[package]]
 name = "embedded-mcu-hal"
 version = "0.1.0"
-source = "git+https://github.com/OpenDevicePartnership/embedded-mcu#1903e774f4ef86735f0c16bf59ca911412ee5534"
+source = "git+https://github.com/OpenDevicePartnership/embedded-mcu#4dadf27b212ce91d884e9a33b3c2725d5d0511fb"
 dependencies = [
  "defmt",
  "num_enum",
@@ -707,7 +707,7 @@ dependencies = [
 [[package]]
 name = "storage_bus"
 version = "0.1.0"
-source = "git+https://github.com/OpenDevicePartnership/embedded-mcu#1903e774f4ef86735f0c16bf59ca911412ee5534"
+source = "git+https://github.com/OpenDevicePartnership/embedded-mcu#4dadf27b212ce91d884e9a33b3c2725d5d0511fb"
 
 [[package]]
 name = "strsim"
--- a/examples/rt633/Cargo.lock
+++ b/examples/rt633/Cargo.lock
@@ -404,7 +404,7 @@ dependencies = [
 [[package]]
 name = "embedded-mcu-hal"
 version = "0.1.0"
-source = "git+https://github.com/OpenDevicePartnership/embedded-mcu#1903e774f4ef86735f0c16bf59ca911412ee5534"
+source = "git+https://github.com/OpenDevicePartnership/embedded-mcu#4dadf27b212ce91d884e9a33b3c2725d5d0511fb"
 dependencies = [
  "defmt",
  "num_enum",
@@ -795,7 +795,7 @@ checksum = "a8f112729512f8e442d81f95a8a7ddf2b7c6b8a1a6f509a95864142b30cab2d3"
 [[package]]
 name = "storage_bus"
 version = "0.1.0"
-source = "git+https://github.com/OpenDevicePartnership/embedded-mcu#1903e774f4ef86735f0c16bf59ca911412ee5534"
+source = "git+https://github.com/OpenDevicePartnership/embedded-mcu#4dadf27b212ce91d884e9a33b3c2725d5d0511fb"
 
 [[package]]
 name = "strsim"
--- a/examples/rt685s-evk/Cargo.lock
+++ b/examples/rt685s-evk/Cargo.lock
@@ -488,7 +488,7 @@ dependencies = [
 [[package]]
 name = "embedded-mcu-hal"
 version = "0.1.0"
-source = "git+https://github.com/OpenDevicePartnership/embedded-mcu#1903e774f4ef86735f0c16bf59ca911412ee5534"
+source = "git+https://github.com/OpenDevicePartnership/embedded-mcu#4dadf27b212ce91d884e9a33b3c2725d5d0511fb"
 dependencies = [
  "chrono",
  "defmt",
@@ -938,7 +938,7 @@ dependencies = [
 [[package]]
 name = "storage_bus"
 version = "0.1.0"
-source = "git+https://github.com/OpenDevicePartnership/embedded-mcu#1903e774f4ef86735f0c16bf59ca911412ee5534"
+source = "git+https://github.com/OpenDevicePartnership/embedded-mcu#4dadf27b212ce91d884e9a33b3c2725d5d0511fb"
 
 [[package]]
 name = "strsim"
--- a/examples/rt685s-evk/README.md
+++ b/examples/rt685s-evk/README.md
@@ -8,11 +8,11 @@ These examples illustrates how to use the embassy-imxrt HAL.
 Add uniquely named example to `src/bin` like `adc.rs`
 
 ## Build
-`cd` to examples folder
+`cd` to examples/[device] folder
 `cargo build --bin <example_name>` for example, `cargo build --bin adc`
 
 ## Run
 Assuming RT685 is powered and connected to Jlink debug probe and the latest probe-rs is installed via  
   `$ cargo install probe-rs-tools --git https://github.com/probe-rs/probe-rs --locked`  
-`cd` to examples folder  
+`cd` to examples/[device] folder  
 `cargo run --bin <example_name>` for example, `cargo run --bin adc`
--- a/examples/rt685s-evk/src/bin/ctimer-pwm-0-100.rs
+++ b/examples/rt685s-evk/src/bin/ctimer-pwm-0-100.rs
@@ -1,0 +1,44 @@
+#![no_std]
+#![no_main]
+
+use defmt::info;
+use embassy_executor::Spawner;
+use embassy_imxrt::gpio;
+use embassy_imxrt::pwm::{CentiPercent, MicroSeconds, Pwm};
+use embassy_imxrt::timer::{CTimerPwm, CTimerPwmPeriodChannel};
+use embassy_time::{Duration, with_timeout};
+use {defmt_rtt as _, embassy_imxrt_examples as _, panic_probe as _};
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
+    let p = embassy_imxrt::init(Default::default());
+
+    info!("PWM test: SCTimer/CTimer based");
+
+    let ctimerperiodchannel = CTimerPwmPeriodChannel::new(p.CTIMER4_COUNT_CHANNEL0, MicroSeconds(1_000)).unwrap();
+
+    let mut monitor = gpio::Input::new(p.PIO1_0, gpio::Pull::None, gpio::Inverter::Disabled);
+
+    let mut ctimer = CTimerPwm::new(p.CTIMER4_COUNT_CHANNEL2, &ctimerperiodchannel, p.PIO0_30).unwrap();
+
+    ctimer.enable(());
+
+    // By default, duty cycle should be 0%, always low, so no rising edges
+    let res = with_timeout(Duration::from_secs(2), monitor.wait_for_rising_edge()).await;
+
+    match res {
+        Err(_) => info!("No rising edge detected as expected for 0% duty cycle"),
+        Ok(_) => panic!("Unexpected rising edge detected at 0% duty cycle"),
+    }
+
+    let duty = CentiPercent(100, 0);
+    ctimer.set_duty((), duty);
+
+    // With 100% duty cycle, always high, so no falling edges
+    let res = with_timeout(Duration::from_secs(2), monitor.wait_for_falling_edge()).await;
+
+    match res {
+        Err(_) => info!("No falling edge detected as expected for 100% duty cycle"),
+        Ok(_) => panic!("Unexpected falling edge detected at 100% duty cycle"),
+    }
+}
--- a/examples/rt685s-evk/src/bin/rtc-alarm.rs
+++ b/examples/rt685s-evk/src/bin/rtc-alarm.rs
@@ -1,0 +1,93 @@
+#![no_std]
+#![no_main]
+
+use core::task::Poll;
+
+use defmt::info;
+use embassy_executor::Spawner;
+use embassy_imxrt::rtc::{Rtc, RtcDatetimeClock};
+use embassy_time::Timer;
+use embedded_mcu_hal::time::{Datetime, DatetimeClock, DatetimeClockError, Month, UncheckedDatetime};
+use {defmt_rtt as _, embassy_imxrt_examples as _, panic_probe as _};
+
+/// RTC alarm struct to await the time alarm wakeup location
+/// This should be implemented by the user to handle RTC peripheral synchronization
+struct RtcAlarm<'r> {
+    expires_at: u64,
+    rtc: &'r RtcDatetimeClock<'r>,
+}
+
+impl<'r> Future for RtcAlarm<'r> {
+    type Output = Result<(), DatetimeClockError>;
+    fn poll(self: core::pin::Pin<&mut Self>, cx: &mut core::task::Context<'_>) -> Poll<Self::Output> {
+        match self.rtc.get_current_datetime() {
+            Ok(now) => {
+                if self.expires_at <= now.to_unix_time_seconds() {
+                    Poll::Ready(Ok(()))
+                } else {
+                    info!("Alarm pending at time {}, expires at {}", now, self.expires_at);
+
+                    // Register our waker to be called by the interrupt handler
+                    self.rtc.register_alarm_waker(cx.waker());
+
+                    Poll::Pending
+                }
+            }
+            Err(e) => Poll::Ready(Err(e)),
+        }
+    }
+}
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) {
+    const ALARM_SECONDS: u64 = 10;
+
+    let p = embassy_imxrt::init(Default::default());
+    let mut r = Rtc::new(p.RTC);
+    let (dt_clock, _rtc_nvram) = r.split();
+
+    // Initialize the system RTC
+    let datetime = Datetime::new(UncheckedDatetime {
+        year: 2026,
+        month: Month::January,
+        day: 12,
+        hour: 16,
+        ..Default::default()
+    })
+    .unwrap();
+    let ret = dt_clock.set_current_datetime(&datetime);
+    info!("RTC set time: {:?}", datetime);
+    assert!(ret.is_ok());
+
+    // Show RTC functioning: Display current time before setting alarm
+    let current_time = dt_clock.get_current_datetime().unwrap();
+    info!("Current time before alarm: {:?}", current_time);
+
+    info!("Waiting 5 seconds...");
+    Timer::after_secs(5).await; // This timer uses the OsTimer peripheral
+
+    info!(
+        "Current time after waiting: {:?}",
+        dt_clock.get_current_datetime().unwrap()
+    );
+
+    // Set an RTC alarm to trigger after ALARM_SECONDS
+    info!("Setting alarm to trigger in {} seconds...", ALARM_SECONDS);
+    let expires_at_secs = dt_clock.set_alarm_from_now(ALARM_SECONDS).expect("Failed to set alarm");
+
+    let alarm = RtcAlarm {
+        expires_at: expires_at_secs,
+        rtc: dt_clock,
+    };
+
+    // Wait for the alarm to trigger
+    info!("Waiting for alarm...");
+    alarm.await.expect("Alarm failed");
+
+    // Display time after alarm triggered
+    let wake_time = dt_clock.get_current_datetime().unwrap();
+    info!("Alarm triggered! Wake time: {:?}", wake_time);
+
+    // Clear the alarm
+    dt_clock.clear_alarm();
+    info!("Alarm cleared");
+}
--- a/src/rtc.rs
+++ b/src/rtc.rs
@@ -2,10 +2,12 @@
 
 use core::marker::PhantomData;
 
+use embassy_hal_internal::interrupt::InterruptExt;
+use embassy_sync::waitqueue::AtomicWaker;
 use embedded_mcu_hal::time::{Datetime, DatetimeClock, DatetimeClockError};
 use embedded_mcu_hal::{Nvram, NvramStorage};
 
-use crate::{Peri, pac, peripherals};
+use crate::{Peri, interrupt, pac, peripherals};
 
 /// Number of general-purpose registers in the RTC NVRAM
 // If you need to consume some of these registers for internal HAL use, write a feature flag that reduces this count,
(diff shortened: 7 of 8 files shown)


src/uart.rs
```diff
@@ -219,11 +220,27 @@ impl<'a> UartTx<'a, Blocking> {
...
 
+struct BufferConfig {
+    #[cfg(feature = "time")]
+    buffer: &'static mut [u8],
+    #[cfg(feature = "time")]
+    write_index: usize,
+    #[cfg(feature = "time")]
+    read_index: usize,
+    #[cfg(feature = "time")]
+    polling_rate: u64,
```
felipebalbi: why do you need `time` feature?

src/uart.rs
```diff
@@ -758,6 +847,163 @@ impl<'a> UartRx<'a, Async> {
...
+                    // copy slices
+                    let incoming_slice = &buffer_config.buffer[in_start..in_stop];
+                    let outgoing_slice = &mut buf[out_start..out_stop];
+                    outgoing_slice.copy_from_slice(incoming_slice);
+
+                    // save off last actual read index in case of longer transfer than expected
+                    buffer_config.read_index = (buffer_config.read_index + read_len) % buffer_config.buffer.len();
+
+                    // finally increment total bytes read
+                    bytes_read += read_len;
```
felipebalbi: look at other `BufferedUart` from other [HALs](https://github.com/embassy-rs/embassy/blob/main/embassy-rp/src/uart/buffered.rs). You'll `embassy_hal_internal` has a ready-made `RingBuffer` for these applications.

felipebalbi (dismissed):

src/uart.rs
```diff
@@ -758,6 +847,163 @@ impl<'a> UartRx<'a, Async> {
...
+    #[cfg(feature = "time")]
+    async fn read_buffered(&mut self, buf: &mut [u8]) -> Result<()> {
+        // unwrap safe here as only entry path to API requires rx_dma instance
+        let rx_dma = self._rx_dma.as_ref().unwrap();
+        let buffer_config = self._buffer_config.as_mut().unwrap();
+
+        // do not need to break into dma lengthed chunks as we already have a reserved circular buffer
+        let mut bytes_read = 0;
+
+        // As the Rx Idle interrupt is not present for this processor, we must poll to see if new data is available
```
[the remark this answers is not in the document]
Neil-alter: Updated from Result<()> to Result<usize>

src/dma/channel.rs
```diff
@@ -137,8 +119,19 @@ impl<'d> Channel<'d> {
...
+                w.clrtrig().clear_bit();
+            } else {
+                w.reload().disabled();
+                w.clrtrig().set_bit();
+            }
+            if options.is_sw_trig {
+                w.swtrig().set_bit();
+            } else {
+                w.swtrig().clear_bit();
+            }
```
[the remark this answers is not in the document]
Neil-alter: Updated

src/uart.rs
```diff
@@ -758,6 +847,163 @@ impl<'a> UartRx<'a, Async> {
...
+                    // save off last actual read index in case of longer transfer than expected
+                    buffer_config.read_index = (buffer_config.read_index + read_len) % buffer_config.buffer.len();
+
+                    // finally increment total bytes read
+                    bytes_read += read_len;
+                    // if we have more bytes to read, do the second part
+                    // only need to copy second part if there's data remaining at the begining of the buffer and we
+                    // still have more bytes requested from read() awaiter
+                    if bytes_read < buf.len() && buffer_config.write_index > 0 {
+                        // do the second part (begining of buffer up to write index)
```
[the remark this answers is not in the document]
Neil-alter: Updated

src/uart.rs
```diff
@@ -757,6 +846,163 @@ impl<'a> UartRx<'a, Async> {
...
+                        let outgoing_slice = &mut buf[out_start..out_stop];
+                        outgoing_slice.copy_from_slice(incoming_slice);
+
+                        // save off last actual read index in case of longer transfer than expected
+                        buffer_config.read_index = (buffer_config.read_index + read_len) % buffer_config.buffer.len();
+
+                        // finally increment total bytes read
+                        bytes_read += read_len;
+                    }
+                }
```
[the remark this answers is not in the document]
Neil-alter: Updated with BBBuffe .

src/uart.rs
```diff
@@ -758,6 +847,163 @@ impl<'a> UartRx<'a, Async> {
...
```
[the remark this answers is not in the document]
Neil-alter: Updated with BBBuffer.

src/uart.rs
Neil-alter: Yes, updated this logic with BBBuffer.

src/dma/transfer.rs
```diff
@@ -6,6 +6,26 @@ use core::task::{Context, Poll};
...
+/// DMA trigger source
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub enum Trigger {
+    /// Software-triggered transfer
+    Software,
+    /// Hardware-triggered transfer
+    Hardware,
+}
+
```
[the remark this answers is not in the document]
Neil-alter: Updated

src/dma/mod.rs
```diff
@@ -27,7 +27,7 @@ const DMA_CHANNEL_COUNT: usize = 33;
 #[derive(Copy, Clone, Debug)]
 #[repr(C)]
 struct ChannelDescriptor {
-    reserved: u32,
+    reserved_reloadcfg: u32,
```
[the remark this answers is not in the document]
Neil-alter: Yes, updated

src/uart.rs
```diff
@@ -752,11 +857,146 @@ impl<'a> UartRx<'a, Async> {
...
+    #[cfg(feature = "time")]
+    async fn read_buffered(&mut self, buf: &mut [u8]) -> Result<usize> {
+        // Extract polling_rate before entering the loop to avoid borrow conflicts
+        let polling_rate = self._buffer_config.as_ref().unwrap().polling_rate;
+        // Bytes already read into buf
+        let mut bytes_read = 0;
+
+        while bytes_read < buf.len() {
+            // Try to pump data from DMA buffer into BBQueue
+            self.pump_data_into_bb();
```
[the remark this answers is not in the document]
jerrysxie: Wait, are we reloading the same buffer into the DMA engine? There is nothing to hold the DMA off from overwriting the data if we are slow in reading the data. We should at least used 2 buffers for ping pong (we can do that with BBQueue of 2048 bytes of buffer and 2 grants of 1024 bytes) so that we have more room for error because there is nothing holding DMA off from overwriting data that software has not had time to read yet.

jerrysxie (changes requested):

src/uart.rs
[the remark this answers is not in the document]
Neil-alter: Hi @jerrysxie , if we pass the BBQueue grant buffer directly to the DMA, does that imply we need to reconfigure and re-enable the DMA channel every time we switch to a new grant? My concern is that repeatedly disabling/enabling the DMA channel could introduce short windows where DMA is inactive. In those windows, incoming UART data might not be captured by DMA and cause FIFO overflow, especially during continuous big package.

This is also a concern for deep sleep scenarios where frequent DMA  re-enabling could result in DMA not being active when UART data arrives.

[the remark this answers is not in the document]
jerrysxie: Hmmm, it seems like grant from bbqueue will be consumed once it is committed. So it is not the best suited for the way we are doing continuous auto-retriggered DMA operation. It is better suited for one shot DMA where DMA will be re-armed after it is completed after every single operation.

Doing it in a hacky way, we can grab 1024 bytes buffer from the first max write grant to know where buffer start and buffer end to give to the DMA. And then we can use transfer count to grant and commit the exact transfer amount to make it work but that is just hacky and we still have do with wraparound situations.

It seems we need a custom buffer implementation to make it work. However, the timing sensitivity of this implementation still concerns me. We are retriggering DMA without any regard to whether the data in the buffer has been read or not. What happens when the entire buffer is overwritten because we are late in reading the buffer? It seems like we are checking the transfer count to know how many bytes has been transferred or has we wrap-around. What happens if we so late that the transfer count wraps more than once. Wouldn't it make more sense to have a ping-pong buffer where the DMA completion interrupt would retrigger the DMA whenever DMA is completed. Then in the interrupt handler we can check whether unread data will be overwritten.

Neil-alter:
Hi @jerrysxie , thanks for the review. Regarding the concern that the producer may outpace the consumer, I have increased the BBQueue size from 1024 to 2048.

In addition, each pump operation limit the maximum amount of data transferred from the DMA buffer to prevent overwhelming the queue in a single iteration.

If the consumer cannot keep up and the BBQueue has no remaining space, we do not silently drop data, we return a overrun flag, this makes data loss explicit to the user.

src/uart.rs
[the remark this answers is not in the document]
jerrysxie: I am not as concerned about overrunning BBQueue buffer; I am more concerned about the auto-retriggered DMA overwritten unread data from previous DMA transactions. Because the polling frequency is passed in through user input. If the user passed in a long polling period, then the DMA will overwritten the unread data in the raw DMA buffer and the user would receive corrupted data because unread data would be overwritten with fresh dta.

Instead of setting the DMA to auto-retriggerable. We could use a ping-pong buffer configuration of 2 1024 buffers and manually retrigger the DMA transfer in the DMA completion interrupt only if the other DMA buffer has been consumed already. And if the buffer has not been consumed then we can set a flag to return an error to the user instead of returning corrupted data. If we did this, it will be much safer. We should be able to do this with BBQueue of 2048 byte total buffer and use 2 1024 grants or even without BBQueue just keep tracking reading from buffer A or buffer B and where we are in the buffer and set a flag when we consume a buffer. 

Why do we need RXIDLE interrupt matter in this case? In the ping-ping buffer case, DMA would be active except in the brief moment when DMA completion interrupt is triggered and we checked whether the previous buffer has been consumed to retrigger the DMA.

src/dma/channel.rs
```diff
@@ -155,6 +160,102 @@ impl<'d> Channel<'d> {
...
+        descriptor_a.dst_data_end_addr = dstbase_a as u32 + (xfercount * xferwidth) as u32;
+        descriptor_a.nxt_desc_link_addr = descriptor_b as *const _ as u32;
+
+        #[allow(clippy::indexing_slicing)]
+        let ping_pong_status = unsafe { &mut PING_PONG_STATUS[channel] };
+        ping_pong_status.current = PingPongSelector::BufferA;
+        ping_pong_status.buffer_a_status = BufferConsumeStatus::Committed;
+        ping_pong_status.buffer_b_status = BufferConsumeStatus::Committed;
+
+        info!("DMA Ping-Pong Descriptors set up on channel {}", channel);
```
jerrysxie: This trace should be removed.

src/dma/mod.rs
```diff
@@ -50,6 +50,59 @@ static mut DESCRIPTORS: DescriptorBlock = DescriptorBlock {
...
+pub enum PingPongSelector {
+    /// DMA buffer A selected
+    BufferA,
+    /// DMA buffer B selected
+    BufferB,
+}
+
+/// Ping Pong buffer status
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum BufferConsumeStatus {
```
jerrysxie: Not blocking, but this can be simplify to BufferStatus.

examples/rt685s-evk/src/bin/uart-async_with_buffer.rs
```diff
@@ -0,0 +1,70 @@
+#![no_std]
```
jerrysxie: Please renamed these example files with dash instead of underscore to keep names consistent.

jerrysxie (changes requested):
How much testing was done using this code? Reworking the example a bit to output 1024 byte of repeating pattern of 0-255 and make read size different every single iteration. I am seeing a data corruption:
```
6354.921461 [INFO ] Total bytes read: 3686364 (uart_async_with_buffer src/bin/uart-async_with_buffer.rs:26)
6354.924497 [INFO ] buf: [220, 221, 222, 223, 224, 225, 226, 227, 228, 229, 230, 231, 232, 233, 234, 235, 236, 237, 238, 239, 240, 241, 242, 243, 109, 110, 111, 112, 113, 114, 115, 116, 117, 118, 119, 120, 121, 122, 123, 124, 125, 126, 127, 128, 129, 130, 131, 132, 133, 134, 135, 136, 137, 138, 139, 140, 141, 142, 143, 144, 145, 146, 147, 148, 149, 150, 151, 152, 153, 154, 155, 156, 157, 158, 159, 160, 161, 162, 163, 164, 165, 166, 167, 168, 169, 170, 171, 172, 173, 174, 175, 176, 177, 178, 179, 180, 181, 182, 183, 184, 185, 186, 187, 188, 189, 190, 191, 192, 193, 194, 195, 196, 197, 198, 199, 200, 201, 202, 203, 204, 205, 206, 207, 208, 209, 210, 211, 212, 213, 214, 215, 216, 217, 218, 219, 135] (uart_async_with_buffer src/bin/uart-async_with_buffer.rs:33)
6354.924622 [ERROR] panicked at src\bin\uart-async_with_buffer.rs:35:17:
Data mismatch at byte 3686388: expected 244, got 109 (panic_probe panic-probe-1.0.0/src/lib.rs:104)
```

src/uart.rs
```diff
@@ -684,11 +703,85 @@ impl<'a> UartRx<'a, Async> {
 
         let rx_dma = dma::Dma::reserve_channel(rx_dma);
 
-        Ok(Self::new_inner::<T>(flexcomm, rx_dma))
+        Ok(Self::new_inner::<T>(flexcomm, rx_dma, None))
+    }
+
+    /// Create a new DMA enabled UART which can only receive data, using a buffer to enable continuous DMA reception via a circular buffer.
```
jerrysxie: Comment is not accurate, it is not a circular buffer.

src/uart.rs
```diff
@@ -828,12 +1145,87 @@ impl<'a> Uart<'a, Async> {
...
             tx: UartTx::new_inner::<T>(flexcomm.clone(), tx_dma),
-            rx: UartRx::new_inner::<T>(flexcomm, rx_dma),
+            rx: UartRx::new_inner::<T>(flexcomm, rx_dma, None),
+        })
+    }
+
+    /// Create a new DMA enabled UART with hardware flow control (RTS/CTS) and Rx buffering enabled
+    #[allow(clippy::too_many_arguments)]
+    #[cfg(feature = "time")]
+    pub fn new_async_with_rtscts_buffer<T: Instance>(
```
jerrysxie: Do we need this always DMA enabled buffer if we have hardware pins to hold off? This seems unnecessary.

src/uart.rs
```diff
@@ -752,11 +845,174 @@ impl<'a> UartRx<'a, Async> {
...
+
+    #[cfg(feature = "time")]
+    async fn read_buffered(&mut self, buf: &mut [u8]) -> Result<usize> {
+        let rx_dma = match self._rx_dma.as_ref() {
+            Some(dma) => dma,
+            None => {
+                debug_assert!(false, "RX DMA not configured");
+                // SAFETY: This branch is unreachable because the only entry path
+                // to this API requires rx_dma to be configured.
+                unsafe { core::hint::unreachable_unchecked() }
```
jerrysxie: This should just return an error.

src/uart.rs
```diff
@@ -752,11 +845,174 @@ impl<'a> UartRx<'a, Async> {
...
+            }
+        };
+        let buffer_config = match self._buffer_config.as_mut() {
+            Some(cfg) => cfg,
+            None => {
+                debug_assert!(false, "Buffer config not initialized");
+                // SAFETY: This branch is unreachable because the only entry path
+                // to this API requires a configured buffer.
+                unsafe { core::hint::unreachable_unchecked() }
```
jerrysxie: This should just return an error.

src/uart.rs
```diff
@@ -752,11 +845,174 @@ impl<'a> UartRx<'a, Async> {
...
+                        let src_slice_opt = buffer_config
+                            .buffer_a
+                            .get(buffer_config.read_off..buffer_config.read_off + to_read);
+                        let dst_slice_opt = buf.get_mut(bytes_read..bytes_read + to_read);
+                        match (src_slice_opt, dst_slice_opt) {
+                            (Some(src_slice), Some(dst_slice)) => {
+                                dst_slice.copy_from_slice(src_slice);
+                            }
+                            _ => {
+                                // Unexpected out-of-bounds; stop reading further to avoid panic.
```
jerrysxie: Should return an error.

src/uart.rs
```diff
@@ -752,11 +845,174 @@ impl<'a> UartRx<'a, Async> {
...
+
+        let half_size = buffer_config.buffer_a.len();
+
+        // Total bytes read into user buffer
+        let mut bytes_read = 0;
+
+        // As the Rx Idle interrupt is not present for this processor, we must poll to see if new data is available
+        while bytes_read < buf.len() {
+            // Check if current buffer has been granted (filled by DMA)
+            let cur_buf = buffer_config.current_buffer;
```
jerrysxie: Calling this `consumer_buf` is more clear.

src/uart.rs
```diff
@@ -752,11 +845,174 @@ impl<'a> UartRx<'a, Async> {
...
+                // Current buffer is equal to the granted buffer
+                // The entire buffer is available to read
+                available = half_size - buffer_config.read_off;
+            } else {
+                // Buffer not yet granted - check if DMA is writing to it
+                // We can try to read partial data using xfercount
+                let dma_current_buffer = rx_dma.current_buffer();
+                if dma_current_buffer == cur_buf {
+                    // DMA is writing to the current buffer, allow to use xfer count to determine available data
+                    let remaining = rx_dma.get_xfer_count() as usize + 1;
```
jerrysxie: XFERCOUNT counts down to 0x3FF instead of 0. If we caught it at exactly 0x3FF, that will cause UB as `half_size - remaining` will overflow is half_size is less than 0x3FF. We should check for this edge case:
image

src/uart.rs
Neil-alter: RTS/CTS provides hardware flow control, but the DMA buffer is still necessary because even with RTS/CTS, this small buffer can overflow during high-speed transfers or when there's processing delay.

examples/rt685s-evk/src/bin/uart-async_with_buffer.rs
Neil-alter: Updated

src/uart.rs
Neil-alter: Updated with read error.

src/uart.rs
```diff
@@ -752,11 +845,182 @@ impl<'a> UartRx<'a, Async> {
...
+            } else {
+                // Buffer not yet granted - check if DMA is writing to it
+                // We can try to read partial data using xfercount
+                let dma_current_buffer = rx_dma.current_buffer();
+                if dma_current_buffer == cur_buf {
+                    // DMA is writing to the current buffer, allow to use xfer count to determine available data
+                    let xfercount = rx_dma.get_xfer_count();
+                    let dma_is_active = rx_dma.is_active();
+                    // Channel not active (transfer finished) and value is 0x3FF - nothing to transfer
+                    let remaining = if (xfercount == 0x3FF) && !dma_is_active {
```
jerrysxie: This will not work as the DMA will stay active until DMA is aborted. And both values are not sampled at the same time. We should just do this:

```
// If transfer count is 0x3FF, the transfer might be been just finished after we check buffer status. Just skip this iteration.
if xfercount == 0x3FF {
 continue;
}
```

src/uart.rs
```diff
@@ -752,11 +845,182 @@ impl<'a> UartRx<'a, Async> {
...
+        let buffer_config = match self._buffer_config.as_mut() {
+            Some(cfg) => cfg,
+            None => {
+                debug_assert!(false, "Buffer config not initialized");
+                return Err(Error::Fail);
+            }
+        };
+
+        // Check for ping-pong buffer overrun error from DMA
+        if rx_dma.check_and_clear_overrun_error() {
```
jerrysxie: This and the rxerr in fifostat should be check every iteration. Because if any overrun occurs, then we should return an error and let the client know that there is data loss.

src/uart.rs
```diff
@@ -788,10 +1052,71 @@ impl<'a> Uart<'a, Async> {
...
+        config: Config,
+        buffer: &'static mut [u8],
+        polling_rate_us: u64,
+    ) -> Result<Self> {
+        tx.as_tx();
+        rx.as_rx();
+
+        let tx = tx.into();
+        let rx = rx.into();
+
```
jerrysxie: Interrupt is not enabled for this initialization.

src/uart.rs
```diff
@@ -752,11 +845,182 @@ impl<'a> UartRx<'a, Async> {
...
+                }
+            } else {
+                // No data available, wait for new data to arrive or error condition
+                let res = select(
+                    embassy_time::Timer::after_micros(buffer_config.polling_rate),
+                    // detect bus errors
+                    poll_fn(|cx| {
+                        self.info.waker.register(cx.waker());
+
+                        self.info.regs.intenset().write(|w| {
```
jerrysxie: This error handling should be clean up like PR #543.

jerrysxie (changes requested):

src/uart.rs
```diff
@@ -828,12 +1153,87 @@ impl<'a> Uart<'a, Async> {
...
+            Some(rts.into()),
+            Some(cts.into()),
+            config,
+        )?;
+
+        if !buffer.len().is_multiple_of(2) {
+            return Err(Error::InvalidArgument);
+        }
+
+        let (buffer_a, buffer_b) = buffer.split_at_mut(buffer.len() / 2);
```
JamesHuard: do you need to enforce that the buffer length is a multiple of 2/4/etc.? Here and elsewhere

JamesHuard (commented):
In terms of strategy, here's my thoughts.

We've tested this PR in its current form and it does solve the performance issues we're facing. It also introduces a few new issues, particularly around power consumption.

There's overall design/constructive problems. A few have been called out directly in comments -- there's probably a few more worth discussing too.

There's quite a few code quality improvements that can be made on top of the design considerations, or even in isolation.

The only flat-out logic error that stands out to me is the one that Jerry caught: https://github.com/OpenDevicePartnership/embassy-imxrt/pull/525/files#r2696858157

Interestingly, we haven't hit this case. 

Here's my suggestion-- let's fix the logic error, then merge this in. Then, we should have at least 4 more follow up PRs or work streams, for the following:

- Address the structural/design in a less time constrained, more design-review setting. In particular, there's a couple of different directions here that warrant at least discussion (new type `BufferedUart` instead of embedding in existing; rework existing DMA Read to allow early exit when bus is "idle" instead of doing polling internally; etc..)
- Address the code quality. There're around 35 comments pertaining to this already. There's a fair bit of complexity that can be removed with some more careful loops/conditionals. There's a lot of state scattered about, making maintenance difficult due to tripping hazards.
- Fix the full duplex UART bug -- the current waker model in Info will cause us to drop the Rx waker if a Tx occurs while Rx is being polled. An easy solution here is to have separate Rx and Tx wakers within the Info struct.
- Fix the power consumption problem introduced by adding constant background UART polling. There's a number of ways we can go about this, leveraging the start bit detection interrupt is one path that's been investigated. Adjusting the design so that we don't have to poll at this level is another.

src/uart.rs
jerrysxie: Yeah, that precondition is checked a few lines above it.

src/uart.rs
```diff
@@ -752,11 +845,182 @@ impl<'a> UartRx<'a, Async> {
...
+                        dma::PingPongSelector::BufferB => dma::PingPongSelector::BufferA,
+                    };
+                }
+            } else {
+                // No data available, wait for new data to arrive or error condition
+                let res = select(
+                    embassy_time::Timer::after_micros(buffer_config.polling_rate),
+                    // detect bus errors
+                    poll_fn(|cx| {
+                        self.info.waker.register(cx.waker());
```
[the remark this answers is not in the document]
jerrysxie: Yeah, previously all error conditions were RX condition so a single waker suffice. If we need to wake due to different conditions, then we will need separate wakers for RX and TX.

huseinmsft (approved):

PR #531, opened 2025-12-17
Audit timers HAL #2: Remove channel array

kurtjd:
We can remove the need for indexing into a channel array by just implementing `From`. We keep the `TimerChannelNum` enum because this can be determined statically, whereas if we kept `channel` as just a `usize` we couldn't get an exhaustive match (and would therefore theoretically be fallible). Tested on hardware. Related: #530 

Going to be one more PR after this to remove the `unwrap`s (keeping that separate since it will likely introduce breaking changes).

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -18,7 +18,7 @@ const CHANNEL_PER_MODULE: usize = 4;
 const PWM_PRECISION_CLK_TICKS_PER_PERIOD: u32 = 500;
 
 /// Enum representing timer channels
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, PartialEq)]
 pub enum TimerChannelNum {
     /// Timer channel 0
     Channel0,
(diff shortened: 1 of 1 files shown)


src/timer.rs
```diff
@@ -30,6 +30,17 @@ pub enum TimerChannelNum {
...
+    fn from(channel: TimerChannelNum) -> Self {
+        match channel {
+            TimerChannelNum::Channel0 => 0,
+            TimerChannelNum::Channel1 => 1,
+            TimerChannelNum::Channel2 => 2,
+            TimerChannelNum::Channel3 => 3,
+        }
+    }
+}
+
```
williampMSFT: there's a crate num_enum in this repo that can generate this implementation for you - you just need to do

```
#[derive(num_enum::IntoPrimitive, num_enum::TryFromPrimitive)]
#[repr(usize)]
```
and then set the variants to whatever you want the numeric values to be

kurtjd:
Closing since #532 captures these changes.

PR #534, opened 2025-12-18
Add input buffer enable/disable support.

GBJin:
Add a new function to set input buffer (enable/disable) for power saving. For Surface products, gpio configured as output pin should has input buffer disabled in deep doze.
image

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -190,6 +190,16 @@ impl<S: Sense> Flex<'_, S> {
             unsafe { w.dirsetp().bits(1 << self.pin.pin()) });
     }
 
+    /// Set the input buffer state
+    ///
+    pub fn set_input_buffer(&mut self, enable: bool) {
+        if enable {
+            self.pin.enable_input_buffer();
+        } else {
+            self.pin.disable_input_buffer();
+        }
+    }
+
     /// Set high
     pub fn set_high(&mut self) {
         self.pin.block().set(self.pin.port()).write(|w|


felipebalbi (approved):
LGTM, but shouldn't every `Output` have input buffer disabled?

kurtjd (approved):

jerrysxie:
> LGTM, but shouldn't every `Output` have input buffer disabled?

`Output` with input buffer disabled cannot read back the state of the pin as an input.

src/gpio.rs
```diff
@@ -190,6 +190,16 @@ impl<S: Sense> Flex<'_, S> {
             unsafe { w.dirsetp().bits(1 << self.pin.pin()) });
     }
 
+    /// Set the input buffer state
+    ///
+    pub fn set_input_buffer(&mut self, enable: bool) {
```
jerrysxie: We can already do this with Sense enabled and disabled state. If we add this API, this would break the typestate.

```
    /// Return a new Flex pin instance with level sensing disabled.
    ///
    /// Consumes less power than a flex pin with sensing enabled.
    #[must_use]
    pub fn disable_sensing(self) -> Flex<'d, SenseDisabled> {
        // Cloning the pin is ok since we consume self immediately
        let new_pin = unsafe { self.pin.clone_unchecked() };
        drop(self);
        Flex::<SenseDisabled>::new(new_pin)
    }
```

jerrysxie (changes requested):

felipebalbi:
> > LGTM, but shouldn't every `Output` have input buffer disabled?
> 
> `Output` with input buffer disabled cannot read back the state of the pin as an input.

when you try to read, grab a `critical_section` then enable input buffer, read, disable input buffer?

jerrysxie:
> > > LGTM, but shouldn't every `Output` have input buffer disabled?
> > 
> > 
> > `Output` with input buffer disabled cannot read back the state of the pin as an input.
> 
> when you try to read, grab a `critical_section` then enable input buffer, read, disable input buffer?

We could do that. Right now sensing is enforced with typestate. `is_high()` and `is_low()` only existed for `Flex<'d, SenseEnabled>`. If we want to make this more dynamic then we have to remove the typestate and returns an error when `is_high()` and `is_low()` are invoked with input buffer disabled.

PR #336, opened 2025-03-05
Add rt685 examples running in RAM

jerrysxie:
Currently, we have to do this to actually run in RAM:

1. `cargo objcopy --release --bin <example_name> -- -O binary <example_name>.bin`
2. `cargo install nxp-header --git https://github.com/OpenDevicePartnership/nxp-header.git`
3. `nxp-header <example_name>.bin`
4. `probe-rs download --chip MIMXRT685SFVKB --binary-format bin <example_name>.bin --base-address 0x08000000`
5. Then reset the EVK to run the program
7. `probe-rs attach --chip MIMXRT685SFVKB target/thumbv8m.main-none-eabihf/release/<example_name>` to see RTT traces

See https://github.com/OpenDevicePartnership/nxp-header/pull/1 for CLI utility to add NXP image header

PR #535, opened 2025-12-18
Audit flexspi HAL for panics

kurtjd:
Removes unwraps and most indexing/slicing (two cases of slicing were left in but are very easy to see they can never panic) from the flexspi HAL.

I use this pattern several times for copying a remainder (which may be less than a 4 byte chunk):

```rust
for (to, from) in remainder.iter_mut().zip(data.to_le_bytes().iter()) {
    *to = *from;
}
```

since it avoids the need to use `copy_from_slice` (with a calculated min length). However, this may be slightly less efficient than the equivalent `copy_from_slice` especially since the compiler should be able to elide the panic path as remainder size can be determined statically (`as_chunks::<N>` guarantees statically that remainder length is less than N). Open to using the slice method with an `allow` lint if folks prefer.

# Testing
The changes in `peripheral.rs` I was able to test by copying the unit tests to another project with my new implementations and verifying they pass (speaking of, I can't figure out how to get `cargo test` to work here).

My other changes I at least tested using the `flexspi-embedded-storage` and `flexspi-nor-flash` examples on the 685evk. Still having issues getting the flexspi-nor-flash example working properly as I'm having trouble receiving defmt logs to verify it's functioning correctly.

Resolves https://github.com/OpenDevicePartnership/embedded-services/issues/643

### DIFF
--- a/examples/rt685s-evk/src/bin/flexspi-storage-service.rs
+++ b/examples/rt685s-evk/src/bin/flexspi-storage-service.rs
@@ -419,7 +419,8 @@ async fn main(_spawner: Spawner) {
             rx_watermark: 0x8,
             tx_watermark: 0x8,
         },
-    );
+    )
+    .unwrap_or_else(|_| panic!("Invalid config"));
 
     // Configure the Flexspi controller
     let _ = flexspi_storage.configport.configure_flexspi(&flexspi_config); // Configure the Flexspi controller
--- a/src/flexspi/nor_storage_bus.rs
+++ b/src/flexspi/nor_storage_bus.rs
@@ -608,7 +608,12 @@ impl BlockingNorStorageBusDriver for FlexspiNorStorageBus<'_, Blocking> {
         read_buf: Option<&mut [u8]>,
         write_buf: Option<&[u8]>,
     ) -> Result<(), NorStorageBusError> {
-        if cmd.data_bytes.is_some() && cmd.data_bytes.unwrap() > MAX_TRANSFER_SIZE_PER_COMMAND as u32 {
+        // `program_lut` expects data_bytes to be available so we bail out early if that's not the case
+        let Some(data_bytes) = cmd.data_bytes else {
+            return Err(NorStorageBusError::StorageBusInternalError);
+        };
+
+        if data_bytes > MAX_TRANSFER_SIZE_PER_COMMAND as u32 {
             return Err(NorStorageBusError::StorageBusInternalError);
         }
 
@@ -616,7 +621,7 @@ impl BlockingNorStorageBusDriver for FlexspiNorStorageBus<'_, Blocking> {
         self.setup_ip_transfer(self.command_sequence_number, cmd.addr, cmd.data_bytes);
 
         // Program the LUT instructions for the command
-        self.program_lut(&cmd, self.command_sequence_number);
+        self.program_lut(&cmd, data_bytes, self.command_sequence_number);
 
         // Start the transfer
         self.execute_ip_cmd();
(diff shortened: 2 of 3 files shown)


jerrysxie:
@kurtjd flexspi-storage-service example is not working? I believe it requires loading the image into RAM. You look at how to do it from this PR: https://github.com/OpenDevicePartnership/embassy-imxrt/pull/336.

src/flexspi/nor_storage_bus.rs
```diff
@@ -1081,14 +1086,24 @@ impl FlexspiNorStorageBus<'_, Blocking> {
                     while self.info.regs.intr().read().iprxwa().bit_is_clear() {}
                 }
             }
-            for (chunk, slot) in watermark_sized_chunk
-                .chunks_mut(FIFO_SLOT_SIZE as usize)
-                .zip(0..num_rx_watermark_slot)
-            {
+
```
jerrysxie: The old code was more straight-forward, please revert and just add a lint exception.

jerrysxie (changes requested):

felipebalbi (dismissed):

src/flexspi/nor_storage_bus.rs
kurtjd: I reverted, added the check to the top of the function, and added a lint exception with panic safety comment. I don't think we should be waiting until this function to check `rx_watermark` but I didn't want to make the constructors which take a config fallible and introduce breaking changes right now.

kurtjd: Put the check in the constructor which introduces a breaking change.

jerrysxie:
> Can you add a testing section to the PR description to indicate which flash example was ran with this change?

Oh, and please announce the breaking change on Zulip, once it is ready to be merged.

src/flexspi/nor_storage_bus.rs
```diff
@@ -1086,6 +1091,10 @@ impl FlexspiNorStorageBus<'_, Blocking> {
                 .zip(0..num_rx_watermark_slot)
             {
                 let data = self.info.regs.rfdr(slot as usize).read().bits();
+                // Panic safety: We've guaranteed a chunk will always be 4 bytes in size,
```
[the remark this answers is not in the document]
felipebalbi: wouldn't, then, `align_to::<u32>()` be more useful here?

madeleyneVaca (approved):

examples/rt685s-evk/src/bin/flexspi-storage-service.rs
```diff
@@ -419,7 +419,8 @@ async fn main(_spawner: Spawner) {
             rx_watermark: 0x8,
             tx_watermark: 0x8,
         },
-    );
+    )
+    .unwrap_or_else(|_| panic!("Invalid config"));
```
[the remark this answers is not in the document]
kurtjd: I think the panic clippy checks don't apply to examples? At least a few of our examples seem to already have unwraps/expects/panics and don't trigger any warnings (since these are allow by default).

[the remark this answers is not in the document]
bramsdell-ms: My mistake -- missed this was an example.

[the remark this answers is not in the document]
kurtjd: Also I used this over a simple `unwrap()` because `unwrap` requires the error type to implement `Debug`, but this just bubbles up an error type from a 3rd party crate which unfortunately does not derive `Debug`.

src/flexspi/nor_storage_bus.rs
```diff
@@ -608,15 +608,20 @@ impl BlockingNorStorageBusDriver for FlexspiNorStorageBus<'_, Blocking> {
         read_buf: Option<&mut [u8]>,
         write_buf: Option<&[u8]>,
     ) -> Result<(), NorStorageBusError> {
-        if cmd.data_bytes.is_some() && cmd.data_bytes.unwrap() > MAX_TRANSFER_SIZE_PER_COMMAND as u32 {
+        // `program_lut` expects data_bytes to be available so we bail out early if that's not the case
+        let Some(data_bytes) = cmd.data_bytes else {
```
bramsdell-ms: Maybe:

`let data_bytes = cmd.data_bytes.ok_or(NorStorageBusError::StorageBusInternalError)?`

src/flexspi/peripheral.rs
```diff
@@ -535,11 +541,17 @@ impl<'a> FlexSpi<'a> {
...
+
+        // If this somehow were to end up out of bounds (maybe from cosmic rays?),
+        // we return 0 bytes written instead of panicking
         let copy_len = buffer.len().min(fill_level).min(watermark);
+        let Some(dest) = buffer.get_mut(..copy_len) else {
+            return 0;
+        };
 
         // Copy data from TX FIFO to buffer.
         unsafe {
```
ayodejiige: Safety comment

PR #537, opened 2026-01-06
Fix incorrect flexspi error handling

kurtjd:
I had mistakenly thought `program_lut` always requires a `Some(data_bytes)` but looking at it again it only needs `Some(data_bytes)` in the case of `Some(transfertype)`.

Because of my mistaken assumption, I was returning an error early from `send_command` when I shouldn't have, so this PR fixes it so we don't do that.


ayodejiige (approved):

tullom (dismissed):
Looks good, just one nonblocking comment

tullom (approved):

PR #544, opened 2026-01-15
imxrt: Update embedded-io to v0.7

kurtjd:
Other HALs have updated to v0.7 for embedded-io, and cargo can't unify the two versions, so consumers (like the uart-service) which expects one version won't work with the other. This doesn't break the API, but if our partners are depending on anything that consumes embedded-io v0.6 this might cause breaking changes.

Note: The updated Error trait requires `core::error::Error` and `Display`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -293,8 +293,8 @@ dependencies = [
  "embedded-hal 1.0.0",
  "embedded-hal-async",
  "embedded-hal-nb",
- "embedded-io",
- "embedded-io-async",
+ "embedded-io 0.7.1",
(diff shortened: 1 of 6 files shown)


williampMSFT (dismissed):

RobertZ2011 (approved):

kurtjd:
Marked as draft since this will likely cause breaking changes for consumers (for example, a partner is implementing `embedded-async-io` for a `SharedUart` which wraps this Uart, but the old version of `embedded-async-io` didn't have `flush()` as a required trait method so that would need to be added).

PR #553, opened 2026-04-07
Update formatting to agree with nightly cargo fmt

williampMSFT:
It looks like there were some changes to the format imposed by `cargo fmt` in the nightly builds, which are checked as part of our CI pipelines.
Update formatting to align with the nightly formatting rules to unblock our CI pipelines.

### DIFF
--- a/examples/rt633/src/bin/espi.rs
+++ b/examples/rt633/src/bin/espi.rs
@@ -4,6 +4,7 @@
 use core::slice;
 
 use defmt::{error, info};
+use defmt_rtt as _;
 use embassy_executor::Spawner;
 use embassy_imxrt::bind_interrupts;
 use embassy_imxrt::espi::{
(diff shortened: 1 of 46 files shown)


kurtjd (approved):

jerrysxie (approved):
@williampMSFT @kurtjd Maybe we should nuke the nightly build checks. Nightly build has been problematic for us at time.

williampMSFT:
> @williampMSFT @kurtjd Maybe we should nuke the nightly build checks. Nightly build has been problematic for us at time.

Yeah, I was wondering why this didn't hit embedded-services; looks like embedded-services does some nightly stuff, but fmt isn't one of them?  What's our motivation for running the nightly checks at all?

kurtjd:
> > @williampMSFT @kurtjd Maybe we should nuke the nightly build checks. Nightly build has been problematic for us at time.
> 
> Yeah, I was wondering why this didn't hit embedded-services; looks like embedded-services does some nightly stuff, but fmt isn't one of them? What's our motivation for running the nightly checks at all?

Yeah I'm not entirely sure what the motivation was for nightly fmt either. I don't see any issue with just replacing it with stable (though rustfmt.toml will need to be updated as well since the config in there only works on nightly).

PR #557, opened 2026-04-29
Bump embassy-time deps to drop heapless 0.8

felipebalbi:
Updates the following embassy-time crates to their latest patches in order to drop heapless 0.8 from the resolved dependency tree:

  - embassy-time:             0.5.0 -> 0.5.1
  - embassy-time-driver:      0.2.1 -> 0.2.2
  - embassy-time-queue-utils: 0.3.0 -> 0.3.2

The 0.3.2 release of embassy-time-queue-utils switched its heapless constraint from ^0.8 to ^0.9, which is what allowed embassy-sync 0.8.0 (already on heapless 0.9.2) and the queue-utils crate to converge on a single heapless major. The other two patch bumps are pulled in by the same lockfile resolution and are kept consistent across the workspace and all three example workspaces.

Lockfile heapless versions after this change:

  - root:                                heapless 0.9.2 only
  - examples/rt633:                      heapless 0.9.2 only
  - examples/rt685s-evk-performance-tracing: heapless 0.9.2 only
  - examples/rt685s-evk:                 heapless 0.8.0 + 0.9.2

The rt685s-evk example still resolves to two heapless majors because keyberon (consumed from git master) and usb-device 0.3.2 (latest on crates.io) both pin heapless ^0.8 transitively. Both are upstream- controlled and out of scope for this commit; rt685s-evk will fully converge once those upstream crates release a heapless 0.9-compatible version.

cargo vet was run and the embassy-time-queue-utils 0.3.0 -> 0.3.2 delta was certified as safe-to-deploy in supply-chain/audits.toml. The other patch bumps did not introduce any new unvetted packages.

Assisted-by: Claude:claude-opus-4.7

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,9 +81,9 @@ mimxrt633s = ["mimxrt633s-pac", "_mimxrt633s"]
 [dependencies]
 storage_bus = { git = "https://github.com/OpenDevicePartnership/embedded-mcu" }
 embassy-sync = "0.8.0"
-embassy-time-driver = { version = "0.2.1", optional = true }
-embassy-time-queue-utils = { version = "0.3.0", optional = true }
-embassy-time = { version = "0.5.0", optional = true }
+embassy-time-driver = { version = "0.2.2", optional = true }
+embassy-time-queue-utils = { version = "0.3.2", optional = true }
+embassy-time = { version = "0.5.1", optional = true }
 embassy-embedded-hal = { version = "0.6.0", default-features = false }
 embassy-futures = "0.1.2"
 embassy-hal-internal = { version = "0.3.0", features = [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -362,7 +362,7 @@ dependencies = [
  "embedded-io-async 0.7.0",
  "futures-core",
  "futures-sink",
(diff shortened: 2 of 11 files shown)


kurtjd (approved):

Cargo.toml
```diff
@@ -81,9 +81,9 @@ mimxrt633s = ["mimxrt633s-pac", "_mimxrt633s"]
 [dependencies]
 storage_bus = { git = "https://github.com/OpenDevicePartnership/embedded-mcu" }
 embassy-sync = "0.8.0"
-embassy-time-driver = { version = "0.2.1", optional = true }
-embassy-time-queue-utils = { version = "0.3.0", optional = true }
-embassy-time = { version = "0.5.0", optional = true }
+embassy-time-driver = { version = "0.2.2", optional = true }
+embassy-time-queue-utils = { version = "0.3.2", optional = true }
+embassy-time = { version = "0.5.1", optional = true }
```
jerrysxie, line 86: Do we need to bubble this up the entire ODP stack? Does the `embedded-service` main still build against `embassy-imxrt` main?

jerrysxie (approved):

PR #565, opened 2026-06-01
i2c::slave: implement embedded-mcu-hal i2c::target trait

felipebalbi:
## Summary

This PR implements the `embedded_mcu_hal::i2c::target` trait on the existing `I2cSlave` driver and adds two on-target examples that exercise it.

The driver gains two trait implementations:

- `embedded_mcu_hal::i2c::target::blocking::I2c` for `I2cSlave<'_, Blocking>`
- `embedded_mcu_hal::i2c::target::asynch::I2c` for `I2cSlave<'_, Async>`

Each is implemented twice — once for `SevenBitAddress` and once for `TenBitAddress` — with a runtime guard that returns `ErrorKind::Other` when the wrong impl is driven against the configured address.

The existing inherent `listen` / `respond_to_write` / `respond_to_read` API and the `Command` / `Response` enums are preserved verbatim. The trait impls are layered on top via private helpers that return richer internal events / terminations, mapped down to either flavour at the boundary. The existing `i2c-slave.rs` and `i2c-slave-async.rs` examples build and behave unchanged.

## What's new

- **`recover()`** as an inherent method on both flavours: aborts in-flight DMA, NAKs any pending byte, disables DMA arming, clears the deselect latch, and drops any pending remediation flag. The configured address(es) and `slven` bit are preserved across recover, matching the trait contract.
- **`PendingEdge` bookkeeping** that surfaces a `Sr` (repeated START) terminator from a previous `respond_to_*` as a `Request::RepeatedStart(prev_addr)` on the next `listen()` call — for trait callers only; the inherent API is unaffected.
- **`Probe → Stop(addr)`** mapping: a zero-byte write surfaces as `Request::Stop(addr)` with no preceding `Write`, documented in the module rustdoc.
- **`Error::kind()`** mapping for every `TransferError` variant + `UnsupportedConfiguration`, with unit tests in `#[cfg(test)] mod tests`.

## Bug fix included

Mid-implementation HIL testing surfaced a pre-existing-style bug in the **new** async DMA read helper that this PR introduces: the `poll_fn` was missing the symmetric counterpart of the write path's `!dma_channel.is_active() && is_slave_receive()` wake condition. When DMA finished draining the supplied buffer but the controller kept clocking SCL for more bytes, the hardware did not raise `slvpending` (it just floated SDA so the controller clocked `0xFF` from the bus pull-ups), and the future slept until something external aborted the transaction.

The fix adds the missing `!dma_channel.is_active() && is_slave_transmit()` wake arm plus the matching post-await `InternalTermination::Continued(buf_len) → ReadStatus::NeedMore` mapping. This is the second commit in the PR for ease of review.

## Examples

Two new rt685s-evk examples (FC2 / `PIO0_18` SCL, `PIO0_17` SDA — same wiring as `i2c-slave-async.rs`, so any existing controller test rig works unchanged):

- `examples/rt685s-evk/src/bin/i2c-slave-async-target-trait.rs`
- `examples/rt685s-evk/src/bin/i2c-slave-target-trait.rs`

Each example wires the trait's `listen → match Request → respond_* → loop` loop and handles every `Request` / `WriteStatus` / `ReadStatus` variant.

## HIL validation

Verified against an RT685S-EVK driven by a Pico de Gallo USB-to-I2C bridge. The full test matrix passes:

- 7-bit address scan finds the slave at `0x20`.
- 8-byte write → `WriteStatus::Stopped(8)`.
- 8-byte read → `ReadStatus::Complete(8)`, payload matches `[0..7]`.
- 4-byte write → `Sr` → 8-byte read in a single transaction → `Restarted(4)` + `RepeatedStart(0x20)` + `Read(0x20)` + `Complete(8)`.
- 24-byte write → three `BufferFull(8)` cycles + `Stopped(0)`.
- 4-byte read of an 8-byte buffer → `EarlyStop(4)`.
- 16-byte read of an 8-byte buffer → `NeedMore(8)` + `Complete(8)` (the bug-fix scenario above).

## Dep bump

`embedded-mcu-hal` is bumped from 0.2.0 to 0.3.0, which is the first release containing the `i2c::target` trait module.

## Other notes

- No new feature flags.
- No changes to register access, `unsafe` blocks, clock setup, or pin muxing.
- Clippy clean on both the `mimxrt685s` and `mimxrt633s` feature combos.
- All existing examples in `examples/rt685s-evk/` still build.

### DIFF
--- a/examples/rt685s-evk-performance-tracing/src/bin/multiple-tasks.rs
+++ b/examples/rt685s-evk-performance-tracing/src/bin/multiple-tasks.rs
@@ -4,6 +4,7 @@
 extern crate embassy_imxrt_perf_examples;
 
 use core::sync::atomic::{AtomicBool, Ordering};
+
 use embassy_executor::Spawner;
 use embassy_imxrt::gpio;
 use embassy_time::Timer;
--- a/src/i2c/slave.rs
+++ b/src/i2c/slave.rs
@@ -811,13 +811,53 @@ impl I2cSlave<'_, Async> {
         let i2c = self.info.regs;
         let buf_len = buf.len();
 
-        // Verify that we are ready for write
+        // Re-entry classification.
+        //
+        // Two shapes reach this entry point: the inherent caller has just
+        // returned from `listen()` with `Command::Write` (the
+        // straight-through case), or it is looping back with a fresh
+        // buffer after a prior call returned `Response::WriteContinued`
+        // (the continuation case). In the continuation case the bus state
+        // may have advanced between the previous call returning and this
+        // call being polled — the controller can issue a `STOP` (slvdesel
+        // latches) or a repeated `START` followed by ADDR+R/W (the
+        // peripheral transitions out of `slave_receive` into
+        // `slave_address`, with `slvpending` asserted waiting for SW to
+        // ACK/NACK the new address).
+        //
+        // Both terminators are legitimate end-of-transaction events from
+        // the slave's perspective. Surface them as the same `Response`
+        // variants the post-poll path emits so the caller's outer loop
+        // sees a uniform shape regardless of when the controller
+        // terminated. The blocking flavour already handles this via its
+        // polling loop; the async flavour must check explicitly because
+        // it arms DMA before awaiting and needs the bus to actually be in
+        // `slave_receive` for the DMA to make forward progress.
         let stat = i2c.stat().read();
+        if stat.slvdesel().is_deselected() {
+            // Controller issued STOP — either between calls (continuation
+            // case) or this is a zero-byte write that the caller
+            // dispatched after a `Probe`-shaped match. Clear the latch
+            // before returning so the next `listen()` does not see a
+            // stale deselect.
+            i2c.stat().write(|w| w.slvdesel().deselected());
+            return Ok(Response::Stopped(0));
+        }
+        if stat.slvstate().is_slave_address() {
+            // Controller issued Sr+ADDR between BufferFull continuations.
+            // The new address phase is latched and waiting for SW
+            // ACK/NACK; the next `listen()` will continue_ it and pick
+            // up the new direction. Queue the RepeatedStart edge for any
+            // trait caller and report Restarted so the outer loop loops
+            // back to `listen()` instead of arming a DMA transfer for a
+            // transaction that no longer exists.
+            self.queue_edge(EdgeKind::RepeatedStart);
+            return Ok(Response::Restarted(0));
+        }
         if !stat.slvstate().is_slave_receive() {
-            // 0 byte write
-            if stat.slvdesel().is_deselected() {
-                return Ok(Response::Stopped(0));
-            }
+            // The peripheral is in some other transient state we do not
+            // know how to drain from here. Surface as a read failure so
+            // the caller can recover().
             return Err(TransferError::ReadFail.into());
         }
 
@@ -897,8 +937,43 @@ impl I2cSlave<'_, Async> {
         let i2c = self.info.regs;
         let buf_len = buf.len();
 
+        // Re-entry classification.
+        //
+        // Two shapes reach this entry point: the inherent caller has just
+        // returned from `listen()` with `Command::Read` (the
+        // straight-through case), or it is looping back with a fresh
+        // buffer after a prior call returned `Response::ReadContinued`
+        // (the continuation case). In the continuation case the bus state
+        // may have advanced between the previous call returning and this
+        // call being polled — the controller can NACK+STOP (slvdesel
+        // latches) or issue a repeated `START` followed by ADDR+R/W (the
+        // peripheral transitions out of `slave_transmit` into
+        // `slave_address`, with `slvpending` asserted waiting for SW to
+        // ACK/NACK the new address).
+        //
+        // Surface both terminators with the same `Response` variants the
+        // post-poll path emits so the caller's outer loop sees a uniform
+        // shape regardless of when the controller terminated. Mirrors the
+        // re-entry classification in `respond_to_write`.
+        let stat = i2c.stat().read();
+        if stat.slvdesel().is_deselected() {
+            // Controller NACKed + STOPped. Clear the latch and report as
+            // `ReadEarlyStop(0)` — no bytes from this call's buffer
+            // reached the wire (the buffer is untouched). The caller can
+            // distinguish "buffer fully consumed at termination" from
+            // "buffer not consumed at termination" via the variant.
+            i2c.stat().write(|w| w.slvdesel().deselected());
+            return Ok(Response::ReadEarlyStop(0));
+        }
+        if stat.slvstate().is_slave_address() {
+            // Controller issued Sr+ADDR between ReadContinued
+            // continuations. Queue the RepeatedStart edge and report
+            // Restarted so the outer loop loops back to `listen()`.
+            self.queue_edge(EdgeKind::RepeatedStart);
+            return Ok(Response::Restarted(0));
+        }
         // Verify that we are ready for transmit
-        if !i2c.stat().read().slvstate().is_slave_transmit() {
+        if !stat.slvstate().is_slave_transmit() {
             return Err(TransferError::WriteFail.into());
         }
 
--- a/src/i2c/slave.rs
+++ b/src/i2c/slave.rs
@@ -1221,10 +1221,12 @@ impl mcu_target::blocking::I2c<SevenBitAddress> for I2cSlave<'_, Blocking> {
     }
 
     fn respond_to_read(&mut self, buf: &[u8]) -> Result<mcu_target::ReadStatus> {
+        self.require_address_kind(false)?;
         Ok(resp_to_read_status(I2cSlave::<Blocking>::respond_to_read(self, buf)?))
     }
 
     fn respond_to_write(&mut self, buf: &mut [u8]) -> Result<mcu_target::WriteStatus> {
+        self.require_address_kind(false)?;
         Ok(resp_to_write_status(I2cSlave::<Blocking>::respond_to_write(self, buf)?))
     }
 }
@@ -1247,10 +1249,12 @@ impl mcu_target::blocking::I2c<TenBitAddress> for I2cSlave<'_, Blocking> {
     }
 
     fn respond_to_read(&mut self, buf: &[u8]) -> Result<mcu_target::ReadStatus> {
+        self.require_address_kind(true)?;
         Ok(resp_to_read_status(I2cSlave::<Blocking>::respond_to_read(self, buf)?))
     }
 
     fn respond_to_write(&mut self, buf: &mut [u8]) -> Result<mcu_target::WriteStatus> {
+        self.require_address_kind(true)?;
         Ok(resp_to_write_status(I2cSlave::<Blocking>::respond_to_write(self, buf)?))
     }
 }
@@ -1273,12 +1277,14 @@ impl mcu_target::asynch::I2c<SevenBitAddress> for I2cSlave<'_, Async> {
     }
 
     async fn respond_to_read(&mut self, buf: &[u8]) -> Result<mcu_target::ReadStatus> {
+        self.require_address_kind(false)?;
         Ok(resp_to_read_status(
             I2cSlave::<Async>::respond_to_read(self, buf).await?,
         ))
     }
 
     async fn respond_to_write(&mut self, buf: &mut [u8]) -> Result<mcu_target::WriteStatus> {
+        self.require_address_kind(false)?;
         Ok(resp_to_write_status(
             I2cSlave::<Async>::respond_to_write(self, buf).await?,
         ))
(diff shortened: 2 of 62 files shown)


kurtjd (commented):
There seems to be quite a bit of code duplication. Looks like you've defined several `*_inner` methods which are mostly copy and paste of the existing methods they mirror with slightly different return types to be used by the trait impls. Is it not possible to just use the exusting methods and just map their output in the trait impls instead of duplicating code?

src/i2c/slave.rs
```diff
@@ -739,3 +846,827 @@ impl Drop for NakGuard {
...
+            }
+            if stat.slvpending().is_pending() {
+                return Poll::Ready(());
+            }
+            // DMA drained but the master still expects more bytes.
+            // Without this guard the future sleeps forever because the
+            // hardware does not raise slvpending once DMA is armed and
+            // running dry — SDA just floats and the master clocks 0xFF
+            // until something else aborts the transaction.
+            if !dma_channel.is_active() && stat.slvstate().is_slave_transmit() {
```
jerrysxie: I had to drop the partial read condition from the original implementation due to a HW race condition, in stress testing, the slv_state() would go to addressed instead of transmitted so the async operation gets stuck:
https://github.com/felipebalbi/embassy-imxrt/commit/62dc59c05b00590d395bfbd74c09125cfe615ebe, and I cannot find a workaround even after working with NXP.

felipebalbi:
> There seems to be quite a bit of code duplication. Looks like you've defined several `*_inner` methods which are mostly copy and paste of the existing methods they mirror with slightly different return types to be used by the trait impls. Is it not possible to just use the exusting methods and just map their output in the trait impls instead of duplicating code?

I'll do it, but it will be a breaking change for the existing API due to `Request` not carrying the address and usize fields we need.

src/i2c/slave.rs
felipebalbi: The change here is on the `read` side, not `write`. Haven't seen this particular race condition. I'll try to add a quick test to my hacky tests (with pico de gallo).

jerrysxie: Yeah, this is in `respond_to_read_internal()`. I was hitting this race condition consistently in long term stressing test with release binary.

src/i2c/slave.rs
```diff
@@ -108,25 +155,139 @@ impl TenBitAddressInfo {
...
+        /// Address that was matched.
+        addr: Address,
+    },
+    /// Controller is about to write to this target. Service it with
+    /// [`I2cSlave::respond_to_write`].
+    Write {
+        /// Address that was matched.
+        addr: Address,
+    },
+}
```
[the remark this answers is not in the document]
felipebalbi: Yeah, that's alright. Crate is unpublished and the goal really was to improve the API.

src/i2c/slave.rs
felipebalbi: unable to hit it. Has been running for quite a while now.

felipebalbi: I'll bring Mole tomorrow for testing, I guess it's really the only way to give this a solid shot. Really cannot reproduce with pico de gallo.

RobertZ2011 (dismissed):

src/i2c/slave.rs
felipebalbi: Mole has been running since 6AM today, no errors.

jerrysxie: Running i2c-loopback-async.rs with this changes to trigger case where i2c target does not provide enough data on single call of respond_to_read() on IMXRT in loopback mode produces a panic:

```
diff --git a/examples/rt685s-evk/src/bin/i2c-loopback-async.rs b/examples/rt685s-evk/src/bin/i2c-loopback-async.rs
index fbdbf50..1086100 100644
--- a/examples/rt685s-evk/src/bin/i2c-loopback-async.rs
+++ b/examples/rt685s-evk/src/bin/i2c-loopback-async.rs
@@ -16,7 +16,7 @@ const ADDR: u8 = 0x20;
const MASTER_BUFLEN: usize = 8;
// slave buffer has to be 1 bigger than master buffer because master does not
// handle end of read properly
-const SLAVE_BUFLEN: usize = MASTER_BUFLEN + 1;
+const SLAVE_BUFLEN: usize = MASTER_BUFLEN / 2;
const SLAVE_ADDR: Option<Address> = Address::new(ADDR);
bind_interrupts!(struct Irqs {
```

```
9140.448752 [INFO ] Read @ "SevenBit(32)" (i2c_loopback_async src/bin/i2c-loopback-async.rs:43)
9140.448841 [INFO ] Response to read got 4 bytes, more bytes to fill (i2c_loopback_async src/bin/i2c-loopback-async.rs:54)
9140.448934 [INFO ] Response to read got 4 bytes, more bytes to fill (i2c_loopback_async src/bin/i2c-loopback-async.rs:54)
9140.485323 [ERROR] panicked at src/bin/i2c-loopback-async.rs:45:68:
called `Result::unwrap()` on an `Err` value: Transfer(WriteFail) (panic_probe panic-probe-1.0.0/src/lib.rs:104)
Firmware exited unexpectedly: Exception
WARN probe_rs_debug::debug_info: UNWIND: Error while checking for exception context. The stack trace will not include the calling frames. : Probe(Register("No Stack Pointer register. Please report this as a bug."))
Core 0
    Frame 0: HardFault_ @ 0x08007a7e
       /home/jxie/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cortex-m-rt-0.7.5/src/lib.rs:1103:1
    Frame 1: UNWIND: Error while checking for exception context. The stack trace will not include the calling frames. : Probe(Register("No Stack Pointer register. Please report this as a bug.")) @ 0x08007a7e
```

The line that triggered this error
```
        // Verify that we are ready for transmit
        if !i2c.stat().read().slvstate().is_slave_transmit() {
            return Err(TransferError::WriteFail.into());
        }
```

This corresponds to the hardware race condition that we engaged with NXP before:

```
We are seeing some inconsistencies on the SLVSTATE behavior:
* In some instance, for I2C writes where slave is in transmit mode. After DMA is completed, we get the DMA interrupt. Master has not NACK any data. SLVSTATE is addressed (0x0) instead of transmit (0x2).
* In some instances, for I2C writes where slave is in transmit mode. After DMA is completed, we get the DMA interrupt. Master is generating NACK and STOP with a little bit of delay. Initially we observed SLVSTATE is set to transmit (2) and SLVSEL is set, then SLVSTATE changes to addressed (0) briefly and SLVSEL is set, then finally SLVSTATE stays as addressed and then SLVDESEL is set and SLVSEL is cleared.
```

We don't have a 100% guarantee that SLVSTATE will be in transmit
```
        poll_fn(|cx| {
            let i2c = self.info.regs;

            self.info.waker.register(cx.waker());
            dma_channel.get_waker().register(cx.waker());

            let stat = i2c.stat().read();
            // Master sent a stop or nack
            if stat.slvdesel().is_deselected() {
                return Poll::Ready(());
            }
            // We need SW intervention
            if stat.slvpending().is_pending() {
                return Poll::Ready(());
            }
            // DMA drained but the master still expects more bytes.
            // Without this guard the future sleeps forever because the
            // hardware does not raise slvpending once DMA is armed and
            // running dry — SDA just floats and the master clocks 0xFF
            // until something else aborts the transaction.
            if !dma_channel.is_active() && stat.slvstate().is_slave_transmit() {  <-- this slvstate is not guaranteed to be in the state we expect
                return Poll::Ready(());
            }

            Poll::Pending
        })
        .await;
```

Maybe we should consider not supporting `Response::ReadContinued(xfer_count)` case on imxrt?

examples/rt685s-evk-performance-tracing/src/bin/multiple-tasks.rs
```diff
@@ -4,6 +4,7 @@
 extern crate embassy_imxrt_perf_examples;
 
 use core::sync::atomic::{AtomicBool, Ordering};
+
```
jerrysxie, line 7: nit: newline?

src/i2c/slave.rs
```diff
@@ -739,3 +1114,310 @@ impl Drop for NakGuard {
...
+    }
+}
+
+/// Map a [`Response`] returned by `respond_to_read` to the trait-side
+/// [`mcu_target::ReadStatus`].
+fn resp_to_read_status(r: Response) -> mcu_target::ReadStatus {
+    match r {
+        Response::ReadComplete(n) => mcu_target::ReadStatus::Complete(n),
+        Response::ReadContinued(n) => mcu_target::ReadStatus::NeedMore(n),
+        Response::ReadEarlyStop(n) | Response::Restarted(n) | Response::Stopped(n) => {
```
jerrysxie, line 1356: There is no possible way for resp_read to returned `Response::Stopped(n)` in the current code, right?

Why is Response::Restarted(n) -> mcu_target::ReadStatus::EarlyStop(n)? We can technically exhaust the DMA and still get a restart? Just trying to understand the terminology here.

src/i2c/slave.rs
```diff
@@ -739,3 +1114,310 @@ impl Drop for NakGuard {
...
+impl mcu_target::blocking::I2c<SevenBitAddress> for I2cSlave<'_, Blocking> {
+    fn recover(&mut self) -> Result<()> {
+        I2cSlave::<Blocking>::recover(self)
+    }
+
+    fn listen(&mut self) -> Result<mcu_target::Request<SevenBitAddress>> {
+        self.require_address_kind(false)?;
+
+        // Drain any queued edge first.
+        if let Some(req) = self.drain_edge::<SevenBitAddress>(cmd_addr_u8) {
```
jerrysxie, line 1375: So we are skipping the listen() step if we restart? So this would be an optimization?

Wouldn't we not skip this transition in listen() to advance the i2c state machine if that is the case?

```
        // Disable DMA
        i2c.slvctl().write(|w| w.slvdma().disabled());

        // Check whether we already have a matched address and just waiting
        // for software ack/nack
        if !i2c.stat().read().slvpending().is_pending() {
            self.poll_sw_action().await;
        }

        if i2c.stat().read().slvstate().is_slave_address() {
            i2c.slvctl().write(|w| w.slvcontinue().continue_());
        } else {
            // If we are already past the addressed phase and in transmit or receive, that means we are already in the
            // next state, most likely due to calling listen() before the previous transaction is completed and leading
            // to state transition out of order. We can tolerate that, so we just move onto the next state.
        }

        // Poll for HW to transitioning from addressed to receive/transmit
        self.poll_sw_action().await;
```

I guess in this case, we left the DMA enabled and the i2c state machine auto transition to the valid state anyway?

jerrysxie: What if after we restarted, we change direction? If we skip listen() and check for directions, do we even know that we change direction?

jerrysxie: Oh, I see, for repeated start, the client would have call listen() again to get the direction. Nevermind.

src/i2c/slave.rs
```diff
@@ -632,33 +935,105 @@ impl I2cSlave<'_, Async> {
...
+            // the FC state machine briefly mis-classifies mid-DMA under
+            // stress (the `slv_state -> addressed` race documented on
+            // PR #565 follow-up discussions).
+            //
+            // Treat it as a normal read termination: NAK to settle the
+            // peripheral and report based on `xfer_count`. Do NOT
+            // synthesise a `RepeatedStart` edge here — that would
+            // fabricate a phantom trait-side event with no matching
+            // upstream Sr on the wire.
+            i2c.slvctl().write(|w| w.slvnack().nack());
```
jerrysxie, line 1125: Slave is sending data to the master, so this nack here does not make sense.

Do we ever hit this branch in testing? With DMA enabled, the slv pending bit is usually never set in my observation.

I left the pending flag check in there previously just as a catch-all in case if something strange ever happens.

src/i2c/slave.rs
jerrysxie: Hmmm, this feels a bit weird, is this the expected sequence of events for start -> read -> start -> read 

1. listen() -> Request::Read(addr)
2. respond_to_read() -> Response::Restarted(n) -> ReadStatus::EarlyStop(n)
3. listen() -> Request::RepeatedStart(prev_addr)
4. listen() -> Request::Read(addr)

RobertZ2011 (approved):

williampMSFT (approved):

jerrysxie:
@felipebalbi This is a breaking changes, so please announce it on Zulip.

Did we ever resolve the subtle race condition? I believe we were planning to return an error and recover from it.

jerrysxie:
> @felipebalbi This is a breaking changes, so please announce it on Zulip.
> 
> Did we ever resolve the subtle race condition? I believe we were planning to return an error and recover from it.

@williampMSFT if there is still HID examples in embedded-services, they would probably be broken after this PR.

williampMSFT:
@jerrysxie 

> @felipebalbi This is a breaking changes, so please announce it on Zulip.
> 
> Did we ever resolve the subtle race condition? I believe we were planning to return an error and recover from it.

if you're talking about the race on mcxa, yeah, apparently one of the other embassy-mcxa maintainers put in a change that incidentally fixed it in a way that doesn't require changes to the trait; I rebased and tested my mock keyboard/mouse on mcxa and couldn't repro the hang anymore (was previously pretty trivial to repro).

re: HID examples, I have a couple new ones that I've written for the new hid-i2c service, will have a PR out for that later today; just need to clean up some //todos I had in there that said "we'll need to rewrite this once the i2c trait rework comes in" and backport some stuff from my device aggregation branch that came up when developing that. Don't have the keyboard service rewrite done yet but if needed I can comment some of that out until I get that sorted.

